Issue #6, opened 2017-10-11
Endpoint::createFromGlobals() spills warnings on PHP-CLI


beberlei:
Accessing `$_SERVER['SERVER_SOFTWARE']` depends on the SAPI, on php cli it is empty.

The code should use `php_sapi_name()` if the server software key is not set.

Additionally, on nginx+fpm the `SERVER_SOFTWARE` says nginx, but that is wrong as well. It should maybe always put the SAPI in there, because the webserver is actually another layer.

jcchavezs:
Great feedback, @beberlei. I will fix it.

jcchavezs:
Closed by https://github.com/jcchavezs/zipkin-php/pull/12

jcchavezs:
Closed by https://github.com/jcchavezs/zipkin-php/pull/12

PR #12, opened 2017-10-15
[#6] Fixes warnings when creating and endpoint from globals in a CLI.

jcchavezs:
Ping @beberlei

### DIFF
--- a/src/Zipkin/Endpoint.php
+++ b/src/Zipkin/Endpoint.php
@@ -70,17 +70,16 @@ class Endpoint
         return new self($serviceName, $ipv4, $ipv6, (int) $port);
     }
 
+    /**
+     * @return Endpoint
+     */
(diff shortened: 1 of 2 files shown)


src/Zipkin/Endpoint.php
```diff
@@ -70,17 +70,16 @@ public static function create($serviceName, $ipv4, $ipv6 = null, $port = null)
         return new self($serviceName, $ipv4, $ipv6, (int) $port);
     }
 
+    /**
+     * @return Endpoint
+     */
     public static function createFromGlobals()
     {
-        if (empty($_SERVER)) {
```
jcchavezs: I removed this as my initial idea was to avoid this from `CLI` but now that I am thinking still makes sense to keep the method.

PR #20, opened 2017-10-25
Makes timestamp optional on Span::finish.

jcchavezs:
For the sake of convenience, the `$timestamp` parameter becomes optional, making it `Timestamp\now()` its default value same as `brave`.

Ping @beberlei

### DIFF
--- a/src/Zipkin/Annotation.php
+++ b/src/Zipkin/Annotation.php
@@ -30,7 +30,7 @@ final class Annotation
 
     /**
      * @param string $value
-     * @param float $timestamp
+     * @param int $timestamp
      * @throws InvalidArgumentException on empty or not stringable value or invalid timestamp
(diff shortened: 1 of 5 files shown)


src/Zipkin/Annotation.php
```diff
@@ -30,7 +30,7 @@ private function __construct($value, $timestamp)
 
     /**
      * @param string $value
-     * @param float $timestamp
+     * @param int $timestamp
```
adriancole: sorry php knowledge.. is int 64bit?

jcchavezs: Good question:

> The INT type in PHP is linked to the CPU and the version of PHP installed (you can have 32bit PHP installed on 64bit CPU).

source: https://stackoverflow.com/a/12049755

PR #24, opened 2017-11-02
Fixes readme

jcchavezs:
Ping @adriancole @beberlei

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,24 +7,22 @@
 
 This is a **production ready** PHP library for Zipkin.
 
+This is a simple example of usage, for a more complete frontend/backend 
+example, check [this repository](https://github.com/openzipkin/zipkin-php-example).
+
 ## Installation
 
 ```bash
 composer require jcchavezs/zipkin
 ```
 
-## Example usage
-
-This is a simple example of usage, for a more complete frontend/backend 
-example, check [this repository](https://github.com/openzipkin/zipkin-php-example).
+## Setup
 
 ```php
 use GuzzleHttp\Client;
 use Zipkin\Annotation;
 use Zipkin\Endpoint;
-use Zipkin\Propagation\DefaultSamplingFlags;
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -45,31 +43,283 @@ $tracing = TracingBuilder::create()
...
+  $span->finish();
+}
+```
+
+In the above example, the span is the root of the trace. In many cases,
+you will be a part of an existing trace. When this is the case, call
+`newChild` instead of `newTrace`
+
+```php
+$span = $tracer->newChild($root->getContext())->setName('encode')->start();
```
adriancole: fyi in hindsight I might have written to just say "nextSpan" as users won't need to know if it is a root or not that way

README.md
```diff
@@ -45,31 +43,283 @@ $tracing = TracingBuilder::create()
...
+}
+```
+
+### Customizing spans
+Once you have a span, you can add tags to it, which can be used as lookup
+keys or details. For example, you might add a tag with your runtime
+version.
+
+```php
+$span->tag('clnt/finagle.version', '6.36.0');
```
adriancole: heh maybe a more relevant tag since this is php

README.md
```diff
@@ -45,31 +43,283 @@ $tracing = TracingBuilder::create()
...
+they add tags and events that relate to their role in an RPC operation.
+
+Here's an example of a client span:
+
+```php
+// before you send a request, add metadata that describes the operation
+$span = $tracer->newTrace()->setName('get')->setKind(Kind\CLIENT);
+$span->tag('clnt/finagle.version', '6.36.0');
+$span->tag(Tags\HTTP_PATH, '/api');
+$span->setRemoteEndpoint(Remote::create('backend', 127 << 24 | 1, null, 8080);
```
adriancole: I'd use IP literals not numbers in endpoint as they are easier. Unless you have other reasons to do that

PR #25, opened 2017-11-05
[#21] Removes annotation class.

### LINKED ISSUE #21: Remove Annotation class?
[Annotation](https://github.com/jcchavezs/zipkin-php/blob/master/src/Zipkin/Annotation.php) was intended to be a value object for holding the guards clause for creating an annotation but actually that can be moved to a `namespace function` and the content can be stored as a `array`.

- In performance terms, according to https://3v4l.org/DQ5jG/perf#output and https://3v4l.org/30jdF/perf#tabs a slightly increase of memory when using arrays but nothing significant whereas the times are way better in arrays than with objects (in both PHP7 and PHP 5.6). There is [post](https://gist.github.com/nikic/5015323) from @nikic explaining why objects use less memory than arrays but an span might not have that many annotations.
- From idiomatic perspective, annotations are opaque meaning that we won't ask any question nor use public methods to it so it makes sense to store as an array rather than as an object.

Ping @beberlei @adriancole @felixfbecker

jcchavezs:
Closes #21 

Ping @beberlei


src/Zipkin/Annotations.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Zipkin\Annotations;
+
+const WIRE_SEND = 'ws';
+const WIRE_RECV = 'wr';
+const LOCAL_COMPONENT = 'lc';
```
adriancole: this and below are tag keys. 'error' can also be a word in an annotation (substring)

jcchavezs: So `LOCAL_COMPONENT` and `ERROR` are tags, right?

adriancole:
yeap as are many others like "http.url" etc.

PR #30, opened 2017-11-16
[#29] Adds support for CurrentTraceContext in tracer.

### LINKED ISSUE #29: Add support for in-process context propagation

jcchavezs:
This PR adds support for `CurrentTraceContext`.

Ping @adriancole @beberlei @felixbecker @cc5092
Closes #29

### DIFF
--- a/src/Zipkin/Propagation/CurrentTraceContext.php
+++ b/src/Zipkin/Propagation/CurrentTraceContext.php
@@ -3,8 +3,7 @@
 namespace Zipkin\Propagation;
 
 /**
- * This makes a given span the current span by placing it in scope (usually but not always a thread
- * local scope).
+ * This makes a given span the current span by placing it in scope.
  *
  * <p>This type is an SPI, and intended to be used by implementors looking to change thread-local
  * storage, or integrate with other contexts such as logging (MDC).
@@ -24,13 +23,19 @@ final class CurrentTraceContext
         $this->context = $currentContext;
     }
 
+    /**
+     * Creates a current trace context controller. If there is no context, and consequently no span, it holds null.
+     *
+     * @param TraceContext|null $currentContext
+     * @return CurrentTraceContext
+     */
     public static function create(TraceContext $currentContext = null)
     {
         return new self($currentContext);
     }
 
     /**
-     * Returns the current span in scope or null if there isn't one.
+     * Returns the current span context in scope or null if there isn't one.
      *
      * @return TraceContext|null
      */
--- a/src/Zipkin/Propagation/CurrentTraceContext.php
+++ b/src/Zipkin/Propagation/CurrentTraceContext.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace Zipkin\Propagation;
+
+/**
+ * This makes a given span the current span by placing it in scope (usually but not always a thread
+ * local scope).
+ *
+ * <p>This type is an SPI, and intended to be used by implementors looking to change thread-local
+ * storage, or integrate with other contexts such as logging (MDC).
+ *
+ * <h3>Design</h3>
+ *
+ * This design was inspired by com.google.instrumentation.trace.ContextUtils,
+ * com.google.inject.servlet.RequestScoper and com.github.kristofa.brave.CurrentSpan
+ */
+
+final class CurrentTraceContext
+{
+    private $context;
+
+    private function __construct(TraceContext $currentContext = null)
+    {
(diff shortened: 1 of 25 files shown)


src/Zipkin/Propagation/CurrentTraceContext.php
```diff
@@ -0,0 +1,59 @@
...
+final class CurrentTraceContext
+{
+    private $context;
+
+    private function __construct(TraceContext $currentContext = null)
+    {
+        $this->context = $currentContext;
+    }
+
+    public static function create(TraceContext $currentContext = null)
```
adriancole: maybe describe what this is for? I guess this is to control the root context for when there's no span in scope?

src/Zipkin/Propagation/CurrentTraceContext.php
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * Sets the current span in scope until the returned object is closed. It is a programming
+     * error to drop or never close the result. Using try-with-resources is preferred for this reason.
+     *
+     * @param TraceContext|null $currentContext
+     * @return callable
+     */
+    public function createScopeAndRetrieveItsCloser(TraceContext $currentContext = null)
```
adriancole: very explicitly named.. I guess that's because you can't declare the type of the result?

src/Zipkin/Tracer.php
```diff
@@ -81,7 +92,7 @@ public function newTrace(SamplingFlags $samplingFlags = null)
     public function newChild(TraceContext $parent)
     {
         if ($parent->isSampled()) {
-            return $this->ensureSampled($this->nextContext($parent));
+            return $this->ensureSampled($this->newRootContext($parent));
```
adriancole: this doesn't look right? newRootContext when we have a parent? maybe make sure you have tests for this to ensure even if looks odd is correct.

src/Zipkin/Tracer.php
```diff
@@ -115,33 +126,101 @@ public function joinSpan(TraceContext $context)
...
+     * <p>Note: While downstream code might affect the span, calling this method, and calling the closer on
+     * the result have no effect on the input. For example, calling closer on the result does not
+     * finish the span. Not only is it safe to call the closer, you must call the closer to end the scope, or
+     * risk leaking resources associated with the scope.
+     *
+     * @param $span span to place into scope or null to clear the scope
+     *
+     * @return callable The scope closer
+     */
+    public function setSpanInScope(Span $span = null)
```
adriancole: no good ideas, but does the word "set" imply one should look at the result? Would be nice to encourage a naming convention that is concise and implies what you should expect even if people skip reading docs. Not sure "with" is necessarily better, just sayin

src/Zipkin/Propagation/CurrentTraceContext.php
jcchavezs: So creating a class for the Scope was adding some overhead just for one method (`close`). Returning a callable is good enough but yeah, I needed to express that it is a closer.

src/Zipkin/Tracer.php
jcchavezs: I was unsure whether to use `set` or `put`. Problem with `with` is that in idiomatic PHP it implies that you are retrieving a copy of the object with the passed value. Maybe `put` is good enough.

src/Zipkin/Propagation/CurrentTraceContext.php
jcchavezs: Sounds reasonable.

src/Zipkin/Tracer.php
beberlei: @jcchavezs `openSpanInScope` maybe? put is ok as well. I agree about idoamtic assumption about with, that is not a good name.

adriancole:
open is a good word as we are opening a scope. maybe openScope(Span).. I
mean do we need to name the arg? This allows us to dodge the naming problem
of how to describe span which isn't the thing being opened :P

src/Zipkin/Tracer.php
jcchavezs: I like `openSpanInScope` very much. WDYT @adriancole ?

jcchavezs:
@icyxp have a look at this PR. Can we have a 5 minutes chat about the changes?

jcchavezs:
@cc5092 can you please have a look at this PR? Can we chat about this changes?

PR #32, opened 2017-11-21
wrong parameter usage. 

cc5092:
spanId will set by parentId and it will be null when it is root span

### DIFF
--- a/src/Zipkin/TraceContext.php
+++ b/src/Zipkin/TraceContext.php
@@ -190,7 +190,7 @@ final class TraceContext implements SamplingFlags
      */
     public function checkTraceId128bits($traceId128bits = false)
     {
-        if($traceId128bits) {
+        if ($traceId128bits) {
             $this->traceId128bits = $traceId128bits;
             $this->traceId = self::traceId($traceId128bits);
         }
@@ -198,7 +198,7 @@ final class TraceContext implements SamplingFlags
 
     private static function traceId($traceId128bits = false)
     {
-        if($traceId128bits) {
+        if ($traceId128bits) {
             return bin2hex(openssl_random_pseudo_bytes(16));
         }
         return self::nextId();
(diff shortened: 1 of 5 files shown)


src/Zipkin/TraceContext.php
```diff
@@ -162,13 +176,34 @@ public function withSampled($isSampled)
...
             $this->isDebug
         );
     }
 
+    /**
+     * check traceId128bits flag
+     * change trace id to 128bits if necessary
+     * @param bool $traceId128bits
+     */
+    public function checkTraceId128bits($traceId128bits = false)
```
jcchavezs: This method worries me for two reasons: 
1. Makes the TraceContext mutable (which I'd discourage)
2. It receives one parameter but only does something if the parameter is `true`, I'd go for a semantic method like `useTraceId128bits()` but again I'd avoid mutation methods.

What do you think?

src/Zipkin/TraceContext.php
```diff
@@ -33,13 +33,19 @@
      */
     private $parentId;
 
-    private function __construct($traceId, $spanId, $parentId, $isSampled, $debug)
+    /**
+     * @var bool
+     */
+    private $traceId128bits;
```
jcchavezs: What do you think about calling this `useTraceId128bits`? IMO that is more descriptive and works better as flag.

src/Zipkin/TraceContext.php
cc5092: $tracer->newTrace() function calls TraceContext::createAsRoot() to create a root span context. 

TraceContext::createAsRoot() will create a 64bits id for trace id and span id.

createAsRoot() has no parameters to know if 128bits trace id is needed

jcchavezs: Yeah, that is true, we could have it like `createAsRoot(SamplingFlags $samplingFlags = null, $useTraceId128bits = false)` and that parameter coming from the tracer. What do you think?

cc5092: I think it's ok. I will update my pr.

cc5092: @jcchavezs updated~

jcchavezs:
Awesome work @cc5092! 🥇

PR #34, opened 2017-11-22
Create CODE_OF_CONDUCT.md

jcchavezs:
Creates a code of conduct


jcchavezs:
That makes sense. This is actually something github provides almost automatically and I wanted to give a try but I agree is way to early to add something like this.

PR #45, opened 2017-12-29
[#27] Removes dependency on guzzle.

### LINKED ISSUE #27: Replace guzzle with curl in the http reporter
Since the http reporting is a very simple & straightforward task there is not need to include the full the guzzle dependency for such an action. A curl implementation could be easily shipped for HTTP reporter. There are some pros and cons:

**Pros:**
- Removes the Guzzle dependency

**Cons:**
- Need to wrap the curl function calls with an interface that could be mocked. We already mock the client for unit test.

**One very important concern is whether we keep the intention to unit test it or to do some integration tests.**

Ping @basvanbeek @beberlei @felixfbecker @adriancole

jcchavezs:
This PR removes the dependency on the guzzle client using `curl` by default. The factory interface allows users to create their own clients tho.

Ping @beberlei @felixfbecker @cc5092 

Closes #27


README.md
```diff
@@ -20,21 +20,19 @@ composer require jcchavezs/zipkin
...
 
 $endpoint = Endpoint::createFromGlobals();
-$client = new Client();
 
 // Logger to stdout
 $logger = new \Monolog\Logger('log');
 $logger->pushHandler(new \Monolog\Handler\ErrorLogHandler());
 
-$reporter = new HttpLogging($client, $logger);
+$reporter = new Http();
```
jcchavezs: :( this is a breaking change but we want to keep consistency with other Zipkin libraries.

PR #47, opened 2018-01-09
Adds cURL integration test.

jcchavezs:
This PR adds an integration test for the cURL HTTP reporter.

Ping @beberlei @felixfbecker


adriancole:
Sorry mildly unrelated, but is the curl variant async?

jcchavezs:
@adriancole nope, we use curl now as we wanted to get rid of the `guzzle http` dependency but this is sync.

PR #58, opened 2018-01-31
Removes try/catch in HTTP reporter.

jcchavezs:
Currently we catch any reporting exception and drop a log record for it but following most common approach we should propagate the exception and let the client decide what to do with it (ex. ignore the error, send a metric, etc.). This PR is not a breaking change but could break clients not doing `try/catch` when reporting.

Ping @adriancole @cc5092 @felixfbecker


adriancole:
Sgtm

PR #60, opened 2018-02-03
Adds tests for timestamp.

jcchavezs:
This PR adds tests for timestamp.


adriancole:
Curious.. how is it possible for a timestamp to be invalid? Is it because
the clock,is pulggable?

jcchavezs:
PHP 5.6 is not typed so when starting, finishing or adding a log to a span you can pass a invalid timestamp and since we store the value as integer it could be negative or bigger than the epoch time for the next 100 years. That is why we validate it.

adriancole:
Thx! Nice to know

PR #67, opened 2018-03-06
Adds support for reporter metrics.

### LINKED ISSUE #64: Add ReporterMetrics
When reporting there is a possibility of failure (failure in network, zipkin server not available, etc). We need a way to deal with those failures without affecting the system operations.

`zipkin-reporter-java` provides a `ReporterMetrics` to deal with those failures: https://github.com/openzipkin/zipkin-reporter-java/blob/master/core/src/main/java/zipkin2/reporter/ReporterMetrics.java

Ping @adriancole

jcchavezs:
This PR adds the interface for reporter metrics. Reporter metrics is the contract for handling with reporting failures (for example: server not available, failure over network, etc.)

The reason for https://github.com/jcchavezs/zipkin-php/compare/adds_support_for_reporter_metrics?expand=1#diff-542892a2dd974b841a279a65f6b8c93aR35 is to keep retro compatibility.

Ping @adriancole @eligijusvitkauskas-home24  @cc5092

Closes #64

### DIFF
--- a/src/Zipkin/Reporters/Http.php
+++ b/src/Zipkin/Reporters/Http.php
@@ -2,9 +2,7 @@
 
 namespace Zipkin\Reporters;
 
-use Exception;
-use Psr\Log\LoggerInterface;
-use Psr\Log\NullLogger;
+use RuntimeException;
 use Zipkin\Recording\Span;
 use Zipkin\Reporter;
 use Zipkin\Reporters\Http\ClientFactory;
@@ -26,12 +24,19 @@ final class Http implements Reporter
      */
     private $options;
 
+    /**
+     * @var Metrics
+     */
+    private $reportMetrics;
+
     public function __construct(
         ClientFactory $requesterFactory = null,
-        array $options = []
+        array $options = [],
+        Metrics $reporterMetrics = null
     ) {
         $this->clientFactory = $requesterFactory ?: CurlFactory::create();
         $this->options = array_merge(self::DEFAULT_OPTIONS, $options);
+        $this->reportMetrics = $reporterMetrics;
     }
 
     /**
@@ -45,6 +50,12 @@ final class Http implements Reporter
         }, $spans));
 
         $client = $this->clientFactory->build($this->options);
-        $client($payload);
+
+        try {
+            $client($payload);
+            $this->reportMetrics->incrementSpans(count($spans));
+        } catch (RuntimeException $e) {
+            $this->reportMetrics->incrementSpansDropped(count($spans));
+        }
     }
 }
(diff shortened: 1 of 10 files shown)


src/Zipkin/Reporters/Http.php
```diff
@@ -45,6 +50,12 @@ public function report(array $spans)
         }, $spans));
 
         $client = $this->clientFactory->build($this->options);
-        $client($payload);
+
+        try {
+            $client($payload);
+            $this->reportMetrics->incrementSpans(count($spans));
```
adriancole: spans is incremented unconditionally.. it is the amount of spans accepted. drop is the amount of spans accepted that were dropped

src/Zipkin/Reporters/Http.php
```diff
@@ -40,11 +45,24 @@ public function __construct(
...
     public function report(array $spans)
     {
+        $this->reportMetrics->incrementSpans(count($spans));
+        $this->reportMetrics->incrementMessages();
+
         $payload = json_encode(array_map(function (Span $span) {
             return $span->toArray();
         }, $spans));
 
+        $payloadLength = strlen($payload);
```
jcchavezs: in PHP, strings are binaries.

src/Zipkin/Reporters/Http.php
```diff
@@ -40,11 +45,24 @@ public function __construct(
      */
     public function report(array $spans)
     {
+        $this->reportMetrics->incrementSpans(count($spans));
+        $this->reportMetrics->incrementMessages();
```
adriancole: I would move this to after you create the payload, as you could error out trying for some weird json reason

src/Zipkin/Reporters/Http.php
```diff
@@ -40,11 +45,24 @@ public function __construct(
...
+
         $payload = json_encode(array_map(function (Span $span) {
             return $span->toArray();
         }, $spans));
 
+        $payloadLength = strlen($payload);
+        $this->reportMetrics->incrementSpanBytes($payloadLength);
+        $this->reportMetrics->incrementMessageBytes($payloadLength);
+
         $client = $this->clientFactory->build($this->options);
```
adriancole: does this client need to be built on each method or can it be shared. I'm nervous this could error

jcchavezs: `curl` call is stateful so yeah, we need to create one every time :(.

src/Zipkin/Reporters/Http.php
jcchavezs: Right, I just moved it.

PR #72, opened 2018-04-18
Adds InMemory reporter.

jcchavezs:
Adds in memory reporter to ease testing of zipkin instrumentations.

Ping @ellisv


jcchavezs:
@ellisv there is no suche a `Reporter::flush` method: https://github.com/jcchavezs/zipkin-php/blob/master/src/Zipkin/Reporter.php#L7

PR #78, opened 2018-06-22
Change SpanMap::getHash() to crc32() to reduce hash collisions

Aramack:
Change \Zipkin\Recording\SpanMap::getHash() to crc32() to reduce hash collisions.

As discussed in https://github.com/jcchavezs/zipkin-php/issues/76, spl_object_hash() can produce hash collisions when the PHP garbage collectors cleans up destroyed spans.

This changes the getHash() function from spl_object_hash() to a crc32() hash of the spanId, traceId and parentId.

### DIFF
--- a/src/Zipkin/Recording/SpanMap.php
+++ b/src/Zipkin/Recording/SpanMap.php
@@ -85,6 +85,6 @@ final class SpanMap
      */
     private function getHash(TraceContext $context)
     {
-        return spl_object_hash($context);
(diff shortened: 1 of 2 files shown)


jcchavezs:
Great work! Only thing is that as @adriancole [pointed out](https://github.com/jcchavezs/zipkin-php/issues/76#issuecomment-398980244) I would just not consider the parentID for this as spanID are suppose to be unique per trace. What do you think?

jcchavezs:
Welcome @Aramack as first contributor! https://github.com/jcchavezs/zipkin-php/graphs/contributors

PR #79, opened 2018-07-09
Set CURLOPT_RETURNTRANSFER to true 

anensshiya-govinthas-ck:
So that the string value of the transfer is returned rather than outputting it directly.

### DIFF
--- a/tests/Integration/Reporters/Http/CurlFactoryTest.php
+++ b/tests/Integration/Reporters/Http/CurlFactoryTest.php
@@ -80,4 +80,37 @@ final class CurlFactoryTest extends PHPUnit_Framework_TestCase
 
         $curlClient('');
     }
+
+    public function testHttpReportingSilentlySendTraces()
+    {
+        $t = $this;
+
+        $server = HttpTestServer::create(
+            function (RequestInterface $request, ResponseInterface &$response) use ($t) {
+                $t->assertEquals('POST', $request->getMethod());
+                $t->assertEquals('application/json', $request->getHeader('Content-Type')[0]);
+                $response = $response->withStatus(202);
+                $response->getBody()->write('Accepted');
+            }
+        );
+
+        $server->start();
+
+        try {
+            $curlClient = CurlFactory::create()->build([
+                'endpoint_url' => $server->getUrl(),
+            ]);
+
+            ob_start();
+            $curlClient(json_encode([]));
+            $output = ob_get_clean();
+            $this->assertEmpty($output);
+        } catch (Exception $e) {
+            $server->stop();
+
+            $this->fail($e->getMessage());
+        } finally {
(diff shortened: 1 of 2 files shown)


jcchavezs:
Thanks for this! I believe this is a similar issue to this one https://github.com/DataDog/dd-trace-php/pull/20/files. Would you mind to add a similar test?

jcchavezs:
Awesome work @anensshiya-govinthas-ck!

Thank you!

tests/Integration/Reporters/Http/CurlFactoryTest.php
```diff
@@ -80,4 +80,37 @@ public function testHttpReportingFailsDueToUnreachableUrl()
...
+        try {
+            $curlClient = CurlFactory::create()->build([
+                'endpoint_url' => $server->getUrl(),
+            ]);
+
+            ob_start();
+            $curlClient(json_encode([]));
+            $output = ob_get_clean();
+            $this->assertEmpty($output);
+        } catch (Exception $e) {
```
eric-chen-ck: @anensshiya-govinthas-ck Similar to my comments here:
https://github.com/jcchavezs/zipkin-php/pull/81/files#r202151672

I don't think this block is needed.

tests/Integration/Reporters/Http/CurlFactoryTest.php
```diff
@@ -80,4 +80,37 @@ public function testHttpReportingFailsDueToUnreachableUrl()
...
+        $server->start();
+
+        try {
+            $curlClient = CurlFactory::create()->build([
+                'endpoint_url' => $server->getUrl(),
+            ]);
+
+            ob_start();
+            $curlClient(json_encode([]));
+            $output = ob_get_clean();
```
eric-chen-ck: @anensshiya-govinthas-ck I think that ideally, `ob_get_clean()` would happen in a `finally` block because if there's any exception between
```
ob_start()
...
ob_get_clean()
```
then `stdout` might continue to be buffered and this could buffer/silence `stdout` for other tests or for as long as the PHP process is running - potentially silencing information, warning, and errors from the test harness, for example.

jcchavezs: I don't think there will be an exception in this specific test.

jcchavezs: True for this test.

tests/Integration/Reporters/Http/CurlFactoryTest.php
jcchavezs: Agree. Could you please include it in the PR improving the other test?

jcchavezs:
@eric-chen-ck thanks a lot for the feedback!

PR #81, opened 2018-07-12
Option to add additional headers to endpoint request.

anensshiya-govinthas-ck:


src/Zipkin/Reporters/Http/CurlFactory.php
```diff
@@ -38,11 +38,13 @@ public function build(array $options = [])
...
-            curl_setopt($handle, CURLOPT_HTTPHEADER, [
+            curl_setopt($handle, CURLOPT_RETURNTRANSFER, true);
+            $defaultHeaders = [
                 'Content-Type: application/json',
                 'Content-Length: ' . strlen($payload),
-            ]);
-            curl_setopt($handle, CURLOPT_RETURNTRANSFER, true);
+            ];
+            $additionalHeaders = (isset($options['headers']) ? $options['headers'] : []);
+            curl_setopt($handle, CURLOPT_HTTPHEADER, array_merge($defaultHeaders, $additionalHeaders));
```
jcchavezs: I am not entirely sure whether we should merge this way or the other way around. The default headers are not meant to be overriden. What do you think?

src/Zipkin/Reporters/Http/CurlFactory.php
```diff
@@ -38,11 +38,17 @@ public function build(array $options = [])
...
-                'Content-Type: application/json',
-                'Content-Length: ' . strlen($payload),
-            ]);
             curl_setopt($handle, CURLOPT_RETURNTRANSFER, true);
+            $defaultHeaders = [
+                'Content-Type' => 'application/json',
+                'Content-Length' => strlen($payload),
+            ];
+            $additionalHeaders = (isset($options['headers']) ? $options['headers'] : []);
+            $headers = array_merge($additionalHeaders, $defaultHeaders);
```
eric-chen-ck: @anensshiya-govinthas-ck Wouldn't we want to reverse the order of the arrays in `array_merge` here?  From my understanding (http://php.net/manual/en/function.array-merge.php), this order would result in `defaultHeaders` values overriding `additionalHeaders`, when it seems one would want the reverse.

tests/Integration/Reporters/Http/CurlFactoryTest.php
```diff
@@ -41,6 +41,37 @@ function (RequestInterface $request, ResponseInterface &$response) use ($t) {
...
+                'headers' => ['From' => 'user@example.com', 'Content-Type' => 'test'],
+            ]);
+
+            $curlClient(json_encode([]));
+        } catch (Exception $e) {
+            $server->stop();
+
+            $this->fail($e->getMessage());
+        } finally {
+            $server->stop();
```
eric-chen-ck: @anensshiya-govinthas-ck This results in `$server->stop()` being called twice in the case of an `Exception`.  Could it be cleaned up?

src/Zipkin/Reporters/Http/CurlFactory.php
eric-chen-ck: (I see that you had a conversation with @jcchavezs about this previously.)

tests/Integration/Reporters/Http/CurlFactoryTest.php
jcchavezs: Problem here is that $this->fail will throw an unhandled exception which will make the stop to never happen. That is why we need the stop before the fail.

tests/Integration/Reporters/Http/CurlFactoryTest.php
```diff
@@ -41,6 +41,37 @@ function (RequestInterface $request, ResponseInterface &$response) use ($t) {
...
+        $server->start();
+
+        try {
+            $curlClient = CurlFactory::create()->build([
+                'endpoint_url' => $server->getUrl(),
+                'headers' => ['From' => 'user@example.com', 'Content-Type' => 'test'],
+            ]);
+
+            $curlClient(json_encode([]));
+        } catch (Exception $e) {
```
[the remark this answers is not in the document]
jcchavezs: This sounds reasonable. It is important to check that we don't end up with zombie processes for not calling the stop.

tests/Integration/Reporters/Http/CurlFactoryTest.php
```diff
@@ -41,6 +41,37 @@ function (RequestInterface $request, ResponseInterface &$response) use ($t) {
...
     }
 
+    public function testHttpReportingSuccessWithExtraHeader()
+    {
+        $t = $this;
+
+        $server = HttpTestServer::create(
+            function (RequestInterface $request, ResponseInterface &$response) use ($t) {
+                $t->assertEquals('POST', $request->getMethod());
+                $t->assertEquals('application/json', $request->getHeader('Content-Type')[0]);
```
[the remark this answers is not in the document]
jcchavezs: In tests I usually prefer to be explicit when a assertion or input changes rather than hiding it in constants. It helps to make concious decisions.

tests/Integration/Reporters/Http/CurlFactoryTest.php
eric-chen-ck: Hi @jcchavezs,

    I'm not sure I follow why `stop()` won't be called if an unhandled exception is thrown in the `try` block.  Here is my code to illustrate this:
```
$ php -a
Interactive shell

php > function foo() {
php {   try {
php {     echo "curl call\n";
php {     throw new Exception('unhandled');
php {   }
php {   finally {
php {     echo "shutdown still called without catch block\n";
php {   }
php { }
php > foo();
curl call
shutdown still called without catch block

Warning: Uncaught Exception: unhandled in php shell code:4
Stack trace:
#0 php shell code(1): foo()
#1 {main}
  thrown in php shell code on line 4
```

Are you saying the unhandled exception will make the `stop()` never happen because of the test harness or something else I haven't included in my example?

tests/Integration/Reporters/Http/CurlFactoryTest.php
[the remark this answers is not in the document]
eric-chen-ck: I don't think it will - see code sample below.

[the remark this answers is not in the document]
eric-chen-ck: @jcchavezs If I understand correctly, this is one of the 2 changes you would like in a new PR?

tests/Integration/Reporters/Http/CurlFactoryTest.php
eric-chen-ck: @jcchavezs And is this the 2nd of 2 changes you would like in a new PR?

It's not clear to me which changes you want and which changes you don't want.

src/Zipkin/Reporters/Http/CurlFactory.php
jcchavezs: Probably the best here is to call them `$requiredHeaders` instead of `$defaultHeader`.

tests/Integration/Reporters/Http/CurlFactoryTest.php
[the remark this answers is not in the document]
jcchavezs: Yes please @eric-chen-ck.

tests/Integration/Reporters/Http/CurlFactoryTest.php
jcchavezs: `$this->fail` throws an exception inside the `catch` block which won't trigger the finally. I don't think this is a required change.

tests/Integration/Reporters/Http/CurlFactoryTest.php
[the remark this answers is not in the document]
eric-chen-ck: @anensshiya-govinthas-ck I think this `catch` block should be removed.

tests/Integration/Reporters/Http/CurlFactoryTest.php
eric-chen-ck: Agreed, no changed needed here and `server->stop()` will only be called once if `catch` block is removed.

PR #87, opened 2018-07-14
Enhancement: Reference phpunit schema as installed with composer

localheinz:
This PR

* [x] references `phpunit.xsd` as installed with `composer`

💁‍♂️ This way it will always refer to the appropriate schema.


jcchavezs:
Thanks @localheinz!

PR #93, opened 2018-10-05
Allow a timeout value to be set on the cURL Call.

### LINKED ISSUE #94: Supporting Timeout as an Option for Reporter
Since PHP is not a multi-threaded programming language, it's good to support passing a timeout value to the reporter.  Clearly, the timeout for this Reported might need to be different from the general timeout value of PHP/Apache for other calls, so passing it as an `option` can be a good idea.

Aideen-Nasirishargh-CK:
Generally speaking, the Zipkin reporting call should be non-blocking and asynchronous.  But since PHP almost doesn't have either of these properties, it's good to at least be sure that reporting of the Zipkin Trace is not causing a major delay in sampled traces.

Closes #94

### DIFF
--- a/src/Zipkin/Reporters/Http/CurlFactory.php
+++ b/src/Zipkin/Reporters/Http/CurlFactory.php
@@ -50,6 +50,10 @@ final class CurlFactory implements ClientFactory
             }, array_keys($headers), $headers);
             curl_setopt($handle, CURLOPT_HTTPHEADER, $formattedHeaders);
 
+            if (isset($options['timeout'])) {
+                curl_setopt($handle, CURLOPT_TIMEOUT, $options['timeout']);
+            }
+
             if (curl_exec($handle) !== false) {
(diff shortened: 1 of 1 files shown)


Aideen-Nasirishargh-CK:
Should resolve https://github.com/openzipkin/zipkin-php/issues/94.

Aideen-Nasirishargh-CK:
Ping @adriancole @jcchavezs

jcchavezs:
Thanks @Aideen-Nasirishargh-CK!

Only think I'd like to mention is that although we have to report synchronusly, it happens after we served the response to the user so it does not block the request :).

Aideen-Nasirishargh-CK:
Yes, that's true.  But still the process continues and in some end-to-end reporting tools the whole duration of the request looks drastically increased.

Thanks for merging it, Jose! :)

jcchavezs:
Ps. It is José Carlos :).

Den fre. 5. okt. 2018, 19:53 skrev Aideen Nasirishargh <
notifications@github.com>:

PR #103, opened 2019-01-22
Fix: missing right parenthesis

persevereVon:


jcchavezs:
Thanks @persevereVon!

PR #111, opened 2019-04-08
chore: improves API usability.

jcchavezs:
This PR breaks the API (already broken when we moved to PHP 7) and notably we changed the way we build spans:

From
```php
$span = $tracer->newChild($context);
$span->setName('encode');
$span->start(1554988060000);
try {
  doSomethingExpensive();
} finally {
  $span->finish();
}
```

to 

```php
$span = $tracer->startSpan('encode', ['parent' => $context, 'start_time' => 1554988060000]);
try {
  doSomethingExpensive();
} finally {
  $span->finish();
}
```

or if we want to create a child on the current one:

```php
$span = $tracer->startNextSpan('encode', ['start_time' => 1554988060000]);
try {
  doSomethingExpensive();
} finally {
  $span->finish();
}
```

This API is inspired in brave (which zipkin PHP v1 was based on) and zipkin-go which has many usability tweaks. Changes are:

- Spans are meant to be started instead of being created and then started (like zipkin-go does)
- Span methods are fluent.

Ping @adriancole @basvanbeek @ghermeto @telemmaite @anensshiya-govinthas-ck

### DIFF
--- a/src/Zipkin/NoopSpan.php
+++ b/src/Zipkin/NoopSpan.php
@@ -51,20 +51,22 @@ final class NoopSpan implements Span
      * set its name without lock contention.
      *
      * @param int|null $timestamp
-     * @return void
+     * @return self
      */
-    public function start(?int $timestamp = null): void
+    public function start(?int $timestamp = null): Span
     {
+        return $this;
     }
 
     /**
      * Sets the string name for the logical operation this span represents.
      *
      * @param string $name
-     * @return void
+     * @return self
      */
-    public function setName(string $name): void
+    public function setName(string $name): Span
     {
+        return $this;
     }
 
     /**
@@ -73,10 +75,11 @@ final class NoopSpan implements Span
      * and that plus its duration as "ss".
      *
      * @param string $kind
-     * @return void
+     * @return self
      */
-    public function setKind(string $kind): void
+    public function setKind(string $kind): Span
     {
+        return $this;
     }
 
     /**
@@ -87,10 +90,11 @@ final class NoopSpan implements Span
      * @param string $key Name used to lookup spans, such as "your_app.version". See {@link Zipkin\Tags} for
      * standard ones.
      * @param string $value value, cannot be <code>null</code>.
-     * @return void
+     * @return self
      */
-    public function setTag(string $key, string $value): void
+    public function tag(string $key, string $value): Span
     {
+        return $this;
     }
 
     /**
@@ -98,11 +102,12 @@ final class NoopSpan implements Span
      *
      * @param string $value A short tag indicating the event, like "finagle.retry"
      * @param int $timestamp
-     * @return void
+     * @return self
      * @see Annotations
      */
-    public function annotate(string $value, int $timestamp): void
+    public function annotate(string $value, int $timestamp): Span
     {
+        return $this;
     }
 
     /**
@@ -111,10 +116,11 @@ final class NoopSpan implements Span
      * It is often expensive to derive a remote address: always check {@link #isNoop()} first!
      *
      * @param Endpoint $remoteEndpoint
-     * @return void
+     * @return self
      */
-    public function setRemoteEndpoint(Endpoint $remoteEndpoint): void
+    public function setRemoteEndpoint(Endpoint $remoteEndpoint): Span
     {
+        return $this;
     }
 
     /**
@@ -152,20 +158,6 @@ final class NoopSpan implements Span
      * @return void
      */
     public function flush(): void
-    {
-    }
-
-    /**
-     * Tags give your span context for search, viewing and analysis. For example, a key
-     * "your_app.version" would let you lookup spans by version. A tag {@link Zipkin\Tags\SQL_QUERY}
-     * isn't searchable, but it can help in debugging when viewing a trace.
-     *
-     * @param string $key Name used to lookup spans, such as "your_app.version". See {@link Zipkin\Tags} for
-     * standard ones.
-     * @param string $value value, cannot be <code>null</code>.
-     * @return void
-     */
-    public function tag(string $key, string $value): void
     {
     }
 }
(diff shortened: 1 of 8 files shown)


src/Zipkin/NoopSpan.php
```diff
@@ -51,20 +51,22 @@ public function getContext(): TraceContext
      * set its name without lock contention.
      *
      * @param int|null $timestamp
-     * @return void
+     * @return self
      */
-    public function start(?int $timestamp = null): void
+    public function start(?int $timestamp = null): self
```
jcchavezs: Making it fluent might not be necessary if we allow to pass tags on creation and start the span by default.

src/Zipkin/Tracer.php
```diff
@@ -81,28 +81,73 @@ public function __construct(
      * }</pre>
      *
      * @param SamplingFlags $samplingFlags
+     * @param array $options
      * @return Span
      */
-    public function newTrace(SamplingFlags $samplingFlags = null): Span
+    public function startSpan(string $name, ?array $options = []): Span
```
jcchavezs: first new method. Conflates newTrace and newChild.

src/Zipkin/Tracer.php
```diff
@@ -81,28 +81,73 @@ public function __construct(
...
+            $span->setKind($options['kind']);
+        }
+        
+        return $span->setName($name)->start($options['start_time'] ?? null);
     }
 
-    /**
-     * Creates a new span within an existing trace. If there is no existing trace, use {@link
-     * #newTrace()} instead.
+        /**
```
jcchavezs: Don't pay attention to this comments yet.

src/Zipkin/Tracer.php
```diff
@@ -81,28 +81,73 @@ public function __construct(
...
+     * so that data is written to the correct trace.
      *
-     * @param TraceContext $parent
+     * @param SamplingFlags|TraceContext $contextOrFlags
      * @return Span
-     * @throws \RuntimeException
+     * @throws RuntimeException
      */
-    public function newChild(TraceContext $parent): Span
+    public function startNextSpan(string $name, ?array $options = []): Span
```
jcchavezs: replaces `nextSpan`

adriancole:
I presume the started right away thing is for local tracing. functionality for deferred start may still come in useful for some instrumentation, though possibly not in php

jcchavezs:
As for the instrumentation I wrote, I can't recall having this need but
could you please give me some example use cases on this? The only case
comes to my mind is in zipkin-instrumentation-sql where we defer the span
creation based on the output of a method (but we use the start time of when
the method was originally called).

Thanks!

fre. 12. apr. 2019, 00:26 skrev Adrian Cole <notifications@github.com>:

adriancole:
mostly we use span.start deferred because you want the timing to be about the execution of the call. if do things like tagging prior to span.start, you can get an accurate measurement of what the call took vs time to derive tags. Another thing is sometimes there's a deferred start in async instrumentation. For example, the difference between assembling a call and invoking it

src/Zipkin/NoopSpan.php
```diff
@@ -111,10 +116,11 @@ public function annotate(string $value, int $timestamp): void
      * It is often expensive to derive a remote address: always check {@link #isNoop()} first!
      *
      * @param Endpoint $remoteEndpoint
-     * @return void
+     * @return self
      */
-    public function setRemoteEndpoint(Endpoint $remoteEndpoint): void
+    public function setRemoteEndpoint(Endpoint $remoteEndpoint): Span
```
adriancole: maybe remove set as others do?

src/Zipkin/RealSpan.php
```diff
@@ -32,15 +32,15 @@ private function __construct(TraceContext $context, Recorder $recorder)
      * @param Recorder $recorder
      * @return RealSpan
      */
-    public static function create(TraceContext $context, Recorder $recorder): self
+    public static function create(TraceContext $context, Recorder $recorder): Span
```
adriancole: interesting that a utility like this is public? does it need to be? or is it too late to change

src/Zipkin/RealSpan.php
```diff
@@ -64,10 +64,10 @@ public function getContext(): TraceContext
      * set its name without lock contention.
      *
      * @param int $timestamp
-     * @return void
+     * @return self
      * @throws \InvalidArgumentException
      */
-    public function start(?int $timestamp = null): void
```
adriancole: maybe mention if called later this overwrites the timestamp

src/Zipkin/RealSpan.php
```diff
@@ -80,17 +80,19 @@ public function start(?int $timestamp = null): void
...
 
     /**
      * Sets the string name for the logical operation this span represents.
      *
      * @param string $name
-     * @return void
+     * @return self
      */
-    public function setName(string $name): void
+    public function setName(string $name): Span
```
adriancole: you remove setTag for tag earlier.. wondering if you aim to be consistent all over or just above

src/Zipkin/Tracer.php
```diff
@@ -76,55 +70,122 @@ public function __construct(
...
-     *   return $this->tracer->newTrace($flags);
+     *   return $this->tracer->startSpan('request', ['parent' => $flags]);
      * }
      * }</pre>
-     *
-     * @param SamplingFlags $samplingFlags
-     * @return Span
      */
-    public function newTrace(SamplingFlags $samplingFlags = null): Span
+    public function startSpan(string $name, array $options = []): Span
```
adriancole: in brave this is special cased for also applying scope eg startScopedSpan. I'm not sure how important it is to save the call to start. Seems higher value doing more than this to break api. I don't know what features are needed, but the rationale for scoped span was arbitrary tracing. advanced span (ex including deferred start) is for instrumentation and usually any repetition is hidden with shared abstractions like http.

PR #113, opened 2019-07-02
fix if endpoint is null, than service name is invalid

minbaby:
`Endpoint::withServiceName` will return a new instance of Endpoint, so when endpoint is null, service name is invalid.

### DIFF
--- a/src/Zipkin/TracingBuilder.php
+++ b/src/Zipkin/TracingBuilder.php
@@ -152,7 +152,7 @@ class TracingBuilder
         if ($localEndpoint === null) {
             $localEndpoint = Endpoint::createFromGlobals();
             if ($this->localServiceName !== null) {
-                $localEndpoint->withServiceName($this->localServiceName);
+                $localEndpoint = $localEndpoint->withServiceName($this->localServiceName);
             }
         }
 


jcchavezs:
@minbaby thanks for fixing this. I am curious about which version of the library are you using, is it `master` or ` 1.3.2`? Reason for ask is that if you are using `1.3.2` we need to change the base for this PR into https://github.com/openzipkin/zipkin-php/tree/v1. If you are using `master` I would be really happy to do a very first release for `2.0.0-beta` cc @adriancole

minbaby:
v1.3.2

jcchavezs:
Awesome. I will release a new version later today.

jcchavezs:
Done [1.3.3](https://github.com/openzipkin/zipkin-php/releases/tag/1.3.3)

letsfire:
It's still 1.3.2 in composer

jcchavezs:
@letsfire as far as I see in [packagist](https://packagist.org/packages/openzipkin/zipkin) it is 1.3.3:

image

Do you have any trouble requiring version 1.3.3?

letsfire:
Thanks. Now I  can see 1.3.3，yesterday i see only 1.3.2 in packagist and the update time is 2018.
I do not known what's wrong with it. Luckly, it's back to normal today. Thank you again

PR #114, opened 2019-07-03
fix if endpoint is null, than service name is invalid

jcchavezs:


jcchavezs:
Reapplying https://github.com/openzipkin/zipkin-php/pull/113 into master

Issue #118, opened 2019-07-30
http reporter always call zipkin server even without sampler


caozhipan:
hello
I set Sampler Rate=0.01, but qps of zipkin-server didn't fall

then I use tcpdump to catch the tcp packages and found that
![image](https://user-images.githubusercontent.com/5044014/62107378-0780b180-b2da-11e9-8c22-6d8f90389521.png)

finnaly, I found this code 
https://github.com/openzipkin/zipkin-php/blob/master/src/Zipkin/Reporters/Http.php
```php
 /**
     * @param Span[] $spans
     * @return void
     */
    public function report(array $spans): void
    {
        $payload = json_encode(array_map(function (Span $span) {
            return $span->toArray();
        }, $spans));
        if ($payload === false) {
            $this->logger->error(
                sprintf('failed to encode spans with code %d', \json_last_error())
            );
            return;
        }
        $client = $this->clientFactory->build($this->options);
        try {
            $client($payload);
        } catch (RuntimeException $e) {
            $this->logger->error(sprintf('failed to report spans: %s', $e->getMessage()));
        }
    }
```
this function didn't consider that empty $spans

jcchavezs:
Fix is included in the new release: https://github.com/openzipkin/zipkin-php/releases/tag/1.3.4

caozhipan:
@jcchavezs  
Is there any ambiguity in what I said?
In your fix commit, there is just some changes in 'tests/Unit/Reporters/HttpTest.php'  and 'tests/Unit/Reporters/HttpMockFactory.php'

how about the commit : https://github.com/openzipkin/zipkin-php/issues/118

jcchavezs:
Hi @caozhipan my bad.

Currently we have the 1.x.x releases coming form v1 branch and master for the WIP v2. I opened 2 PRs: https://github.com/openzipkin/zipkin-php/pull/120 against `v1` and https://github.com/openzipkin/zipkin-php/pull/119 against `master` but only the master one included the change. Thanks for noticing so quickly so I am preparing a new release.

caozhipan:
@jcchavezs 
great!!!

jcchavezs:
You can try 1.3.4 now: https://github.com/openzipkin/zipkin-php/releases/tag/1.3.4

PR #121, opened 2019-07-30
fix empty spans

caozhipan:
[fix empty spans](https://github.com/openzipkin/zipkin-php/issues/118)


caozhipan:
https://github.com/openzipkin/zipkin-php/issues/118

PR #126, opened 2020-01-07
chore: adds v2 branch to be consumible by composer.

jcchavezs:
This PR adds a branch alias for `v2` branch in order to start consuming that branch over composer.

Ping @alexeyshockov


alexeyshockov:
Thanks.

BTW, I'm not familiar with the development here, but what is the difference between `v2` and `master` branches? If I understand currently, the current stable version 1 is supported in `v1` branch, and the next major release is preparing in `master` branch (version 2). Then why does `v2` branch even exist?

alexeyshockov:
Also for `v2` branch [Packagist has automatically created `v2.x-dev` alias](https://packagist.org/packages/openzipkin/zipkin#v2.0.x-dev), so IMO no need for an additional one

jcchavezs:
That is a very good question. My initial idea was to use master too but
then I realized that every time I released a minor (e.g. 2.3), right after
we have to change the `dev-master` alias into `2.4.x-dev`. In the other
hand after the move into v2 the branch alias won't be needed so I am happy
to change it into `dev-master`.

Issue #115, opened 2019-07-03
Add inSpan method for tracer


jcchavezs:
Sometimes you need to instrument operations that are being performed by legacy code. E.g. http call to a service but instead of using a http standard library like guzzle, it is being done with raw curl. Those times, the effort to actually replace the curl call and use a new shiny and instrumented library like guzzle is already a big effort and it is very error prone.

In those cases, having a `Tracer::inSpan($options, $callable, $args)` method would be helpful as that method is going to do tracing around that legacy call without forcing you to touch the logic (more than passing the right variables into the callable). Does that sound a reasonable use case?

Initial feedback from @adriancole:

> I think there have been similar discussions around this.. sleuth has the ability to do this also py_zipkin. to a degree zipkin-js does (promise tracing)

Ping @adriancole @basvanbeek

jcchavezs:
More info:

- `$options` will allow users to pass span name, tags and potentially a parent context (which might not be needed).
- Any expection being thrown inside the callable will be recorded as a tag error and will be pass up.
- The function will return whatever output comes from the callable.

adriancole:
this type of utility has been discussed many times and has been around in various forms for at least 5 years. OpenCensus is probably not the one to cite for prior art in other words.

Most of the time, this is best to do in such a place that it can be an aspect. Because things like naming the span are not something you want in user code. If it has to be in user code, there's usually integrations for things like annotations:

ex 1: py_zipkin https://github.com/Yelp/py_zipkin#usage-3-log-a-span-inside-an-ongoing-trace
ex 2: sleuth https://cloud.spring.io/spring-cloud-sleuth/spring-cloud-sleuth.html#_creating_new_spans

Some systems which have predefined lambda types also have helper utilities libraries.. of course there's still a problem with naming and tagging policy hence the above paragraph. Anyway zipkin-js has a utility for example, for promises

https://github.com/openzipkin/zipkin-js/tree/master/packages/zipkin#local-tracing

This feature is an alternative to just using normal apis, which force some of the questions to be answered, such as naming policy and exception handling. Ex the normal scoped span in brave:

https://github.com/openzipkin/brave/tree/master/brave#in-process-tracing

adriancole:
[...]

PR #128, opened 2020-01-25
feat: adds inSpan and SpanCustomizer.

jcchavezs:
This PR finally tackles #115 and adds the SpanCustomizer interface, allowing to do further customizations in http tracing.

Most important code: https://github.com/openzipkin/zipkin-php/pull/128/files#diff-ced4309e6e2dfd1e487df57084f07566R233

Ping carefully @adriancole

### DIFF
--- a/src/Zipkin/Tracer.php
+++ b/src/Zipkin/Tracer.php
@@ -228,7 +228,8 @@ final class Tracer
      * @param callable $fn
      * @param array $args
      * @param callable $argsParser with signature function (SpanCustomizer $spanCustomizer, array $args): void
-     * @param callable $resultParser  with signature function (SpanCustomizer $spanCustomizer, ?Throwable $e, mixed $result): void
+     * @param callable $resultParser with signature
+     * function (SpanCustomizer $spanCustomizer, ?Throwable $e, mixed $result): void
      */
     public function inSpan($fn, array $args, ?callable $argsParser = null, ?callable $resultParser = null)
     {
--- a/src/Zipkin/Tracer.php
+++ b/src/Zipkin/Tracer.php
@@ -4,12 +4,16 @@ declare(strict_types=1);
 
 namespace Zipkin;
 
+use BadMethodCallException;
 use RuntimeException;
+use Throwable;
+use Zipkin\Propagation\TraceContext;
+use Zipkin\Propagation\SamplingFlags;
 use Zipkin\Propagation\CurrentTraceContext;
 use Zipkin\Propagation\DefaultSamplingFlags;
-use Zipkin\Propagation\SamplingFlags;
-use Zipkin\Propagation\TraceContext;
 use Zipkin\Sampler;
+use Zipkin\SpanCustomizer;
+use Zipkin\SpanCustomizerShield;
 
 final class Tracer
 {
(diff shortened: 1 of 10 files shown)


src/Zipkin/Tracer.php
```diff
@@ -220,6 +224,49 @@ public function nextSpan(?SamplingFlags $contextOrFlags = null): Span
         throw new RuntimeException('Context or flags for next span is invalid.');
     }
 
+    /**
+     * @param callable $fn
+     * @param array $args
+     * @param callable $before
+     * @param callable $after
+     */
+    public function inSpan($fn, array $args, ?callable $argsParser = null, ?callable $resultParser = null)
```
jcchavezs: This is the important piece.

src/Zipkin/SpanCustomizer.php
```diff
@@ -0,0 +1,45 @@
...
+    public function tag(string $key, string $value): void;
+
+    /**
+     * Associates an event that explains latency with the current system time.
+     *
+     * @param string $value A short tag indicating the event, like "finagle.retry"
+     * @param int $timestamp
+     * @return void
+     * @see Annotations
+     */
```
adriancole: explicit timestamp desired?

src/Zipkin/Tracer.php
```diff
@@ -220,6 +225,81 @@ public function nextSpan(?SamplingFlags $contextOrFlags = null): Span
         throw new RuntimeException('Context or flags for next span is invalid.');
     }
 
+    /**
+     * @param callable $fn
```
adriancole: is this like trace local? maybe add a code example

src/Zipkin/SpanCustomizer.php
jcchavezs: That is actually a very nice catch. It turns out only `RealSpan` was making this optional. I improved this typing and other typings in https://github.com/openzipkin/zipkin-php/pull/130/files#diff-107b1bf5b96774ca3d48e6e084485747R107 and https://github.com/openzipkin/zipkin-php/pull/130/files#diff-554d86e667418ae9f107329f98c58bc8R76

src/Zipkin/Tracer.php
jcchavezs: Sure thing.

jcchavezs: What do you think about this? https://github.com/openzipkin/zipkin-php/pull/128/files#diff-ced4309e6e2dfd1e487df57084f07566R228-R265

src/Zipkin/Tracer.php
```diff
@@ -220,6 +225,122 @@ public function nextSpan(?SamplingFlags $contextOrFlags = null): Span
...
+     * @param array $args
+     * @param string|null $name the name of the span
+     * @param callable|null $argsParser with signature function (SpanCustomizer $span, ?array $args = []): void
+     * @param callable|null $resultParser with signature
+     * function (SpanCustomizer $span, $output = null, ?Throwable $e): void
+     */
+    public function inSpan(
+        callable $fn,
+        array $args = [],
+        ?string $name = null,
```
jcchavezs: One thing to answer here is whether we force the user to pass a span name or we infer it from the callable (like we do it now in https://github.com/openzipkin/zipkin-php/pull/128/files#diff-ced4309e6e2dfd1e487df57084f07566R320)

jcchavezs:
Only thing missing here is to define the context extraction for event loop. I will probably hold that until I finish my work with the current instrumentation for `driftphp`.

Ping @mmoreram

jcchavezs:
This is green: https://travis-ci.org/github/openzipkin/zipkin-php/builds/689046500

PR #135, opened 2020-04-08
fix: Correct inaccurate expression

### LINKED ISSUE #134: Inaccuracy in PercentageSampler
Implementation in `\Zipkin\Samplers\PercentageSampler::isSampled()` is a bit inaccurate:
```php
    public function isSampled($traceId)
    {
        return (mt_rand(0, 99) / 100) <= $this->rate;
    }
```
Given `$this->rate = 0.00` would still sample at 1%, while `$this->rate = 0.99` would already sample at 100%.

It should be sufficient to change operator from `<=` to `<`.

pauci:
Fixes #134

### DIFF
--- a/src/Zipkin/Samplers/PercentageSampler.php
+++ b/src/Zipkin/Samplers/PercentageSampler.php
@@ -39,6 +39,6 @@ final class PercentageSampler implements Sampler
      */
     public function isSampled(string $traceId): bool
     {
-        return (\mt_rand(0, 99) / 100) <= $this->rate;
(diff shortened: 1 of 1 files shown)


jcchavezs:
Thank you!

jcchavezs:
Are you using 1.x? I can do a new release with this change.

pauci:
I'm using 1.3.4, but I made custom sampler already, so I'm ok.

jcchavezs:
Good. Since this is a bug fix I will release 1.3.5 anyways.

PR #141, opened 2020-05-05
Adds span customizer

jcchavezs:
This PR introduces the new concept of `SpanCustomizer` which is a port from brave and it is a needed abstractions for upcoming improvements to the library: inSpan and HTTP client/server PSR implementation.

### DIFF
--- a/src/Zipkin/NoopSpan.php
+++ b/src/Zipkin/NoopSpan.php
@@ -83,7 +83,7 @@ final class NoopSpan implements Span
      * @param string $value
      * @return void
      */
-    public function setTag(string $key, string $value): void
+    public function tag(string $key, string $value): void
     {
     }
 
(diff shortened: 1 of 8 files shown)


src/Zipkin/NoopSpan.php
```diff
@@ -148,18 +148,4 @@ public function finish(int $timestamp = null): void
     public function flush(): void
     {
     }
-
```
jcchavezs, line 151: This was a mistake as `setTag` was not part of the interface.

src/Zipkin/SpanCustomizerShield.php
```diff
@@ -0,0 +1,57 @@
...
+{
+    /**
+     * @var Span
+     */
+    private $delegate;
+
+    /**
+     * @var bool
+     */
+    private $isNotNoop = false;
```
jcchavezs, line 17: Not too straighforward to understand but efficient.

PR #149, opened 2020-05-20
PSR Client instrumentation

jcchavezs:
This is the PR that changes the example to use it: https://github.com/openzipkin/zipkin-php-example/pull/18/files#diff-6ccd0c37f2009a9a6e737e4e30fe5007R30-R32


src/Zipkin/Instrumentation/Http/Client/ClientTracing.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Zipkin\Instrumentation\Http\Client;
+
+use Zipkin\Tracing;
+
+class ClientTracing
```
jcchavezs, line 9: How about this `ClientTracing` @adriancole ?

src/Zipkin/Instrumentation/Http/Client/Client.php
```diff
@@ -0,0 +1,86 @@
...
+     * @var Parser
+     */
+    private $parser;
+
+    /**
+     * @var callable
+     */
+    private $requestSampler;
+
+    public function __construct(
```
jcchavezs, line 42: This is how we create the Client out of the `ClientTracing`.

src/Zipkin/Instrumentation/Http/Client/ClientTracing.php
adriancole: yeah you can certainly split like this. in brave it is one thing (HttpTracing) to make it easier to configure the subsystem, but there's no sharing between client and server config wise (except the tracing component). I don't know how package imports work in php, maybe you can consider HttpClientTracing if there could be conflict.

jcchavezs: I renamed it into `Tracing`. Ideally this would be `use Zipkin\Instrumentation\Http\Client\Tracing` and imported with alias `HttpClientTracing` (`use Zipkin\Instrumentation\Http\Client\Tracing as HttpClientTracing`). I had this concern where the naming of `Zipkin\Tracing` and `Zipkin\Instrumentation\Http\Client\Tracing` would clash but honestly you don't use `Zipkin\Tracing` directly as it ha a builder.

jcchavezs: About "the split" I decided to do so as `HttpClientTracing` and `HttpServerTracing` are different purpose classes and share nothing (as you pointed out).

adriancole:
nice job

PR #151, opened 2020-05-28
feat: adds support for b3 single header.

jcchavezs:
This PR adds support for B3 single header.

### DIFF
--- a/src/Zipkin/Propagation/B3.php
+++ b/src/Zipkin/Propagation/B3.php
@@ -17,37 +17,58 @@ final class B3 implements Propagation
     /**
      * 128 or 64-bit trace ID lower-hex encoded into 32 or 16 characters (required)
      */
-    const TRACE_ID_NAME = 'X-B3-TraceId';
+    private const TRACE_ID_NAME = 'X-B3-TraceId';
     
     /**
      * 64-bit span ID lower-hex encoded into 16 characters (required)
      */
-    const SPAN_ID_NAME = 'X-B3-SpanId';
+    private const SPAN_ID_NAME = 'X-B3-SpanId';
     
     /**
      * 64-bit parent span ID lower-hex encoded into 16 characters (absent on root span)
      */
-    const PARENT_SPAN_ID_NAME = 'X-B3-ParentSpanId';
+    private const PARENT_SPAN_ID_NAME = 'X-B3-ParentSpanId';
     
     /**
      * '1' means report this span to the tracing system, '0' means do not. (absent means defer the
      * decision to the receiver of this header).
      */
-    const SAMPLED_NAME = 'X-B3-Sampled';
+    private const SAMPLED_NAME = 'X-B3-Sampled';
     
     /**
      * '1' implies sampled and is a request to override collection-tier sampling policy.
      */
-    const FLAGS_NAME = 'X-B3-Flags';
+    private const FLAGS_NAME = 'X-B3-Flags';
+
+    private const SINGLE_VALUE_NAME = 'b3';
+
+    private const MULTI_VALUE_NAMES = [
+        self::TRACE_ID_NAME,
+        self::SPAN_ID_NAME,
+        self::PARENT_SPAN_ID_NAME,
+        self::SAMPLED_NAME,
+        self::FLAGS_NAME,
+    ];
+
+    public const INJECT_SINGLE_AND_MULTI = 0;
+    public const INJECT_SINGLE_ONLY = 1;
 
     /**
      * @var LoggerInterface
      */
     private $logger;
 
-    public function __construct(LoggerInterface $logger = null)
-    {
+    /**
+     * @var bool
+     */
+    private $injectMultiValues;
+
+    public function __construct(
+        LoggerInterface $logger = null,
+        int $injectHeadersOption = self::INJECT_SINGLE_AND_MULTI
+    ) {
         $this->logger = $logger ?: new NullLogger();
+        $this->injectMultiValues = $injectHeadersOption !== self::INJECT_SINGLE_ONLY;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/Zipkin/Propagation/B3.php
```diff
@@ -17,51 +17,77 @@ final class B3 implements Propagation
...
-    public function __construct(LoggerInterface $logger = null)
-    {
+    /**
+     * @var bool
+     */
+    private $injectMultiValues;
+
+    public function __construct(
+        LoggerInterface $logger = null,
+        int $injectOption = self::INJECT_OPTION_SINGLE_AND_MULTI
```
jcchavezs, line 79: This options should be carefully curated. Also notice we might need to include $extractOption to only extract single at some point *maybe* (cc @adriancole

src/Zipkin/Propagation/Exceptions/InvalidTraceContextArgument.php
```diff
@@ -22,4 +22,9 @@ public static function forParentSpanId(string $parentId): self
     {
         return new self(\sprintf('Invalid parent span id, got %s', $parentId));
     }
+
+    public static function forSampling(string $value): self
+    {
+        return new self(\sprintf('Invalid sampling bit, got %s, expected 1, 0 or d', $value));
```
adriancole, line 28:
```suggestion
        return new self(\sprintf('Invalid sampling value, got %s, expected 1, 0 or d', $value));
```

src/Zipkin/Propagation/B3.php
adriancole: might consider on the inject side letting people construct the choices like we do here vs bundling into a single constant. https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/B3Propagation.java

jcchavezs: Unfortunately we don't have enum in PHP but I think we can do something cool. In the other hand, am I right to assume that `inject single and multi` is not an option in brave?

src/Zipkin/Propagation/SamplingFlags.php
```diff
@@ -19,6 +19,8 @@ public function isSampled(): ?bool;
      */
     public function isDebug(): bool;
 
+    public function isEmpty(): bool;
```
jcchavezs, line 22: In brave, EMPTY sampling flags is a constant, here we can't do that so the best is to add a method for it (unfortunately). In the other hand, it makes sense to add a method as the current way to check an empty is verify sampling decision is null and debug=false but then you need to make sure it is SamplingFlags but no TraceContext: messy.

src/Zipkin/Propagation/B3.php
```diff
@@ -17,51 +17,118 @@ final class B3 implements Propagation
...
     /**
      * @var LoggerInterface
      */
     private $logger;
 
-    public function __construct(LoggerInterface $logger = null)
-    {
+    public function __construct(
+        LoggerInterface $logger = null,
+        string $injector = self::INJECTOR_MULTI
```
jcchavezs, line 89: @adriancole what do you think about this one?

src/Zipkin/Propagation/SamplingFlags.php
adriancole: extractionresult as a union type would be most ideal

src/Zipkin/Propagation/B3.php
adriancole: you mean how do I feel about specifying one or more by keyword? seems to get the job done.

src/Zipkin/Propagation/B3.php
```diff
@@ -17,77 +20,210 @@ final class B3 implements Propagation
...
+
+    private const NO_KIND = 'no_kind';
+
+    /**
+     * @var array[]
+     */
+    private $injectorsFn = [
+        Kind\CLIENT => [self::INJECT_MULTI, self::INJECT_SINGLE],
+        Kind\PRODUCER => [self::INJECT_SINGLE_NO_PARENT],
+        self::NO_KIND => [self::INJECT_MULTI],
```
jcchavezs, line 99: what do you think about this @adriancole? maybe `default` is better?

adriancole: well in brave we don't inject multiple by default. we only inject multi if client and snp if producer. if we could do it again no_kind would be same as producer. just we didn't decide that initially so stuck until brave 6

jcchavezs: So to make sure I understand, defaults would be:

```
- CLIENT => ['multi']
- PRODUCER => ['single_no_parent']
- default => ['single_no_parent']
```

Since existing instrumentations use `Map` in most cases, which is not a RemoteSetter, then by upgrading they all will start injecting SNP instead of MULTI so a breaking change. How did you fixed that in brave? 

Maybe you did not have the issue that Map is the default getter/setter. This is an issue because PHP did not have standard api for HTTP request/response and hence every framework implemented their own thing. As a consequence we ended up doing weird things in map, for example it is case insensitive which is cumbersome and problematic.

src/Zipkin/Propagation/B3.php
adriancole: in brave there are 3 format keys: SINGLE SINGLE_NO_PARENT MULTI

by default, only one format is configured, but it varies on span kind.

For example,
no span kind is MULTI (but should be SINGLE_NO_PARENT).
kind.client is MULTI
kind.producer is SINGLE_NO_PARENT

you can configure multiple kinds also. for example, you can say client is both MULTI and also SINGLE_NO_PARENT. However, we have logic to prevent SINGLE and SINGLE_NO_PARENT at the same time as that doesn't make sense.

This means the valid combinations are
SINGLE
SINGLE_NO_PARENT
MULTI
SINGLE,MULTI
SINGLE,SINGLE_NO_PARENT

Hope that helps

src/Zipkin/Propagation/B3.php
adriancole: in this case you can make default MULTI like we did in brave until the Map thing ages out

PR #154, opened 2020-06-10
fix(propagation): fixes map lookup for case insensitiviness.

jcchavezs:
Although not ideal, this implementation is more efficient as it only iterates the array until it is needed instead of changing the case for the full array every time.

### DIFF
--- a/src/Zipkin/Propagation/Map.php
+++ b/src/Zipkin/Propagation/Map.php
@@ -34,8 +34,16 @@ final class Map implements Getter, Setter
                 return null;
             }
 
-            $lcCarrier = \array_change_key_case($carrier, \CASE_LOWER);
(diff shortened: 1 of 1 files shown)


src/Zipkin/Propagation/Map.php
```diff
@@ -34,8 +34,13 @@ public function get($carrier, string $key): ?string
                 return null;
             }
 
-            $lcCarrier = \array_change_key_case($carrier, \CASE_LOWER);
-            return \array_key_exists($lKey, $lcCarrier) ? $lcCarrier[$lKey] : null;
```
drolando, line 38: headers are also `-` and `_` insensitive according to the spec

jcchavezs: U_U that makes me sad, however b3 headers use dash, if there is any underscore that is a bug somewhere I think. I will keep it as it is because this "map" is already a mess.

PR #155, opened 2020-06-11
feat(join-spans): adds support for join-spans.

jcchavezs:
This PR adds support for joint spans.

cc @drolando @adriancole

### DIFF
--- a/src/Zipkin/Propagation/Propagation.php
+++ b/src/Zipkin/Propagation/Propagation.php
@@ -64,7 +64,7 @@ interface Propagation
      * identifiers. Other propagation formats, like <a href="https://github.com/w3c/trace-context">trace-context</a>
      * only propagate the calling trace and span ID, with an assumption that the receiver always
      * starts a new child span. When join is supported, you can assume that when {@link
-     * TraceContext#parentId() the parent span ID} is null, you've been propagated a root span. When
+     * TraceContex::getParentId() the parent span ID} is null, you've been propagated a root span. When
      * join is not supported, you must always fork a new child.
      */
     public function supportsJoin(): bool;
--- a/src/Zipkin/Propagation/Propagation.php
+++ b/src/Zipkin/Propagation/Propagation.php
@@ -54,4 +54,18 @@ interface Propagation
      * @return callable
      */
     public function getExtractor(Getter $getter): callable;
+
+    /**
+     * Does the propagation implementation support sharing client and server span IDs. For example,
+     * should an RPC server span share the same identifiers extracted from an incoming request?
+     *
+     * In usual <a href="https://github.com/openzipkin/b3-propagation">B3 Propagation</a>, the
+     * parent span ID is sent across the wire so that the client and server can share the same
+     * identifiers. Other propagation formats, like <a href="https://github.com/w3c/trace-context">trace-context</a>
+     * only propagate the calling trace and span ID, with an assumption that the receiver always
+     * starts a new child span. When join is supported, you can assume that when {@link
+     * TraceContext#parentId() the parent span ID} is null, you've been propagated a root span. When
+     * join is not supported, you must always fork a new child.
+     */
+    public function supportsJoin(): bool;
 }
--- a/src/Zipkin/Propagation/TraceContext.php
+++ b/src/Zipkin/Propagation/TraceContext.php
@@ -39,6 +39,11 @@ final class TraceContext implements SamplingFlags
     /**
      * @var bool
      */
+    private $isShared;
+
+    /**
+     * @var bool
+     */
     private $usesTraceId128bits;
 
     private function __construct(
@@ -47,6 +52,7 @@ final class TraceContext implements SamplingFlags
         ?string $parentId,
         ?bool $isSampled,
         bool $isDebug,
+        bool $isShared,
         bool $usesTraceId128bits
     ) {
         $this->traceId = $traceId;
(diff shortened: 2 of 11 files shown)


src/Zipkin/TracingBuilder.php
```diff
@@ -145,6 +150,12 @@ public function beingNoop(bool $isNoop = true): self
         return $this;
     }
 
+    public function supportingJoin(bool $supportsJoin): self
```
drolando, line 153: This isn't a very self-explanatory name. Could we name it something like `usingSharedSpans`?

jcchavezs: I think `supportJoin` is the name in the other libraries (brave, zipkin-js and zipkin-go). Maybe `supportingJointSpans` is better?

src/Zipkin/Propagation/TraceContext.php
```diff
@@ -222,7 +251,6 @@ public function isEqual(SamplingFlags $samplingFlags): bool
             && $this->spanId === $samplingFlags->spanId
             && $this->parentId === $samplingFlags->parentId
             && $this->isSampled === $samplingFlags->isSampled
-            && $this->isDebug === $samplingFlags->isDebug
-            && $this->usesTraceId128bits === $samplingFlags->usesTraceId128bits;
```
jcchavezs, line 226: This only depends on the traceId length so no point to compare it.

src/Zipkin/Propagation/TraceContext.php
```diff
@@ -222,7 +251,6 @@ public function isEqual(SamplingFlags $samplingFlags): bool
...
+            && $this->isDebug === $samplingFlags->isDebug;
```
jcchavezs, line 254: I am avoiding comparing the `isShared` comparison on purpose. Main reason is because the context is the same, it is just the shared what changed. Also notice this method is mainly for testing.

src/Zipkin/Tracer.php
```diff
@@ -134,7 +141,11 @@ public function newChild(TraceContext $parent): Span
      */
     public function joinSpan(TraceContext $context): Span
     {
-        return $this->toSpan($context);
+        if (!$this->supportsJoin) {
+            return $this->newChild($context);
+        }
+
+        return $this->toSpan($context->withShared(true));
```
jcchavezs, line 148: I am not mutating the context anywhere but returning a new object with same data but marked as shared in here. I believe context does not need to be marked as `shared` everywhere as I don't expect people to create another span out of it (only different between this context and the original one is the shared on span reporting). There are no threads here either.

src/Zipkin/TracingBuilder.php
drolando: Keep it then. Anyway the api doesn't automatically understand if you want to use shared spans and if the current one is shared or not, so we need to call joinSpan anyway

src/Zipkin/Propagation/TraceContext.php
adriancole: we use isShared by default in brave and lens as if you have a loopback it is a different span. in php it might be impossible for the same thing to become a problem as you possibly might not see the client and server side ever at the same time.

src/Zipkin/Recording/SpanMap.php
```diff
@@ -20,13 +20,9 @@ final class SpanMap
...
     {
-        $contextHash = $this->getHash($context);
+        $contextHash = self::getHash($context);
 
-        if (!\array_key_exists($contextHash, $this->map)) {
-            return null;
-        }
-
-        return $this->map[$contextHash];
+        return $this->map[$contextHash] ?? null;
```
adriancole, line 25: ex this hash here is where a clash might happen if loopback is possible. the data recorded against the shared span (server)shouldn't override the pending client side.

src/Zipkin/Tracer.php
adriancole: if there's a mechanism for "withShared" to be internal, I'd use it.. users are unlikely to need to set a context shared..

src/Zipkin/TracingBuilder.php
adriancole: maybe copying docs will help.

src/Zipkin/Recording/SpanMap.php
jcchavezs: Good to know! As you said I don;t think it is possible but who knows what people do so keeping as it is until someone complains.

src/Zipkin/Propagation/Propagation.php
```diff
@@ -54,4 +54,18 @@ public function getInjector(Setter $setter): callable;
...
+    /**
+     * Does the propagation implementation support sharing client and server span IDs. For example,
+     * should an RPC server span share the same identifiers extracted from an incoming request?
+     *
+     * In usual <a href="https://github.com/openzipkin/b3-propagation">B3 Propagation</a>, the
+     * parent span ID is sent across the wire so that the client and server can share the same
+     * identifiers. Other propagation formats, like <a href="https://github.com/w3c/trace-context">trace-context</a>
+     * only propagate the calling trace and span ID, with an assumption that the receiver always
+     * starts a new child span. When join is supported, you can assume that when {@link
+     * TraceContext#parentId() the parent span ID} is null, you've been propagated a root span. When
```
adriancole, line 67: unless php has same doc syntax as java.. fix?

src/Zipkin/Tracer.php
jcchavezs: Unfortunately there isn't. Either it is public for everyone or for none :(. I will make sure I'll add a comment to it.

src/Zipkin/Propagation/Propagation.php
jcchavezs: It is similar but I am not sure this anchor is going to work. I am changing it.

PR #156, opened 2020-06-12
Support extracting incoming headers from the $_SERVER variable

drolando:
If you don't have access to a Request object, it's convenient to extract the headers from the global $_SERVER variable.


jcchavezs:
Thanks @drolando

PR #158, opened 2020-06-13
Allow using custom Recorder and add option to always record spans

drolando:
This will let me implement "firehose mode", similar to what we have in
py_zipkin.

Basically what I need is a way to call 2 different Reporters:
- one gets called only when sampled is True
- one gets called every time (this is the firehose recorder)
This cannot be accomplished with just one Reporter, because once we're
inside the reporter we've lost any info on whether the trace was sampled
or not.

I can almost do this in zipkin-php, there are only 2 things missing.
- if sampled is False, we create a NoopSpan. So I need an extra flag to
  inform the tracer that I always want to create a RealSpan
- the Recorder is hardcoded inside the Tracer class, so I cannot pass my
  own.

Once those 2 things are supported, I can write my own Recorder subclass
that calls the right Reporters as needed with minimal changes to the
core zipkin-php code.

cc @adriancole

### DIFF
--- a/src/Zipkin/Recording/Span.php
+++ b/src/Zipkin/Recording/Span.php
@@ -128,7 +128,7 @@ final class Span
     private $localEndpoint;
 
     /**
-     * @var bool
+     * @var bool|null
      */
     private $isSampled;
 
@@ -139,7 +139,7 @@ final class Span
         bool $debug,
         bool $shared,
         Endpoint $localEndpoint,
-        bool $isSampled = false
+        ?bool $isSampled = false
     ) {
         $this->traceId = $traceId;
         $this->parentId = $parentId;
@@ -238,7 +238,7 @@ final class Span
      */
     public function isSampled(): bool
     {
-        return $this->isSampled;
+        return $this->isSampled === true;
     }
 
     /**
--- a/src/Zipkin/Recording/Span.php
+++ b/src/Zipkin/Recording/Span.php
@@ -127,13 +127,19 @@ final class Span
      */
     private $localEndpoint;
 
+    /**
+     * @var bool|null
+     */
+    private $isSampled;
+
     private function __construct(
         string $traceId,
         ?string $parentId,
         string $spanId,
         bool $debug,
         bool $shared,
-        Endpoint $localEndpoint
+        Endpoint $localEndpoint,
+        ?bool $isSampled = false
     ) {
         $this->traceId = $traceId;
         $this->parentId = $parentId;
@@ -141,6 +147,7 @@ final class Span
         $this->debug = $debug;
         $this->shared = $shared;
         $this->localEndpoint = $localEndpoint;
+        $this->isSampled = $isSampled;
     }
 
     /**
@@ -156,7 +163,8 @@ final class Span
             $context->getSpanId(),
             $context->isDebug(),
             $context->isShared(),
-            $localEndpoint
+            $localEndpoint,
+            $context->isSampled()
         );
     }
 
@@ -223,6 +231,14 @@ final class Span
     public function setRemoteEndpoint(Endpoint $remoteEndpoint): void
     {
         $this->remoteEndpoint = $remoteEndpoint;
+    }
+
+    /**
+     * @return bool
+     */
+    public function isSampled(): bool
+    {
+        return $this->isSampled === true;
     }
 
     /**
--- a/src/Zipkin/Recording/Span.php
+++ b/src/Zipkin/Recording/Span.php
@@ -128,7 +128,7 @@ final class Span
     private $localEndpoint;
 
     /**
-     * @var bool|null
+     * @var bool
      */
     private $isSampled;
 
@@ -139,7 +139,7 @@ final class Span
         bool $debug,
         bool $shared,
         Endpoint $localEndpoint,
-        ?bool $isSampled = false
+        bool $isSampled = false
     ) {
         $this->traceId = $traceId;
         $this->parentId = $parentId;
@@ -238,7 +238,7 @@ final class Span
      */
     public function isSampled(): bool
     {
-        return $this->isSampled === true;
+        return $this->isSampled;
     }
 
     /**
--- a/tests/Unit/TracingBuilderTest.php
+++ b/tests/Unit/TracingBuilderTest.php
@@ -5,6 +5,7 @@ namespace ZipkinTests\Unit;
 use Zipkin\Tracing;
 use Zipkin\Endpoint;
 use Zipkin\Reporters\Noop;
+use Zipkin\Reporters\InMemory;
 use Zipkin\TracingBuilder;
 use Zipkin\Propagation\Getter;
 use Zipkin\Propagation\Setter;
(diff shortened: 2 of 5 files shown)


adriancole:
in Brave we solve this with sampledLocal internal flag

https://github.com/openzipkin/brave/blob/169d93ac71fce759fb872636576d539c6c3b5123/brave/src/main/java/brave/propagation/SamplingFlags.java#L89-L102

then the zipkin "recorder" would look at if it is supposed to act as default (only send ones remote sampled), or wildcard send (alwaysSampleLocal)
https://github.com/openzipkin/zipkin-reporter-java/blob/master/brave/src/main/java/zipkin2/reporter/brave/ZipkinSpanHandler.java#L147


Note: something slightly different what you are doing vs alwaysSampleLocal used in secondary sampling.. in secondary sampling how netflix use it, they add  a comma-separated tag "sampled_keys" so that their firehose proxy can decide what to do vs decide in-process. Ex. if that tag includes "b3" they know it was sampled because of b3 vs wildcard or some overlay.

https://github.com/openzipkin-contrib/zipkin-secondary-sampling/blob/master/docs/design.md#the-sampled_keys-tag
https://github.com/openzipkin-contrib/zipkin-secondary-sampling/blob/master/brave/src/test/java/brave/secondary_sampling/BasicUsageTest.java#L86-L90

If making the decision in-process you don't need to add a tag, just route it to another sender based on logic you deploy.

drolando:
Doing something similar here would mean adding a new `sampledLocal` flag to `TraceContext` for each span. Then in the reporter I can loop through them and send all the ones that have `sampledLocal=true` to firehose, correct?

drolando:
And I see you have the same `alwaysReportSpans` flag too

adriancole:
yeap

sampledlocal is true when sampled is true or if alwaysReportSpans is set. (later you can get fancy and conditionally set sampledLocal based on http headers)

alwaysReportSpans is a tracer-scoped flag where sampledLocal is a span/context scoped flag

jcchavezs:
@drolando thanks for this. I am reluctant to expose the Recorder to the user, mostly because I always saw it as an internal type and making it customizable would create some weird API surface (starting by the fact that we pass the reporter as a parameter already and it is supposed to be included in the recorder aswell). I'd prefer the approach @adriancole suggests by using the sampledLocal flag.

@adriancole I got a question, to summarize what we could do is:

0. Include the Context Flags or at leas the sampledLocal in the MutableSpan (so the reporter has access to sampledLocal, not sure which option is the best)
1. Create the SpanHandler interface whose default class only forwards the remote sampled spans (notice this can enable better things like only forward remote sampled spans with an error or certain duration)
2. Let the Reporter to decide how to multiplex the spans based on the sampledLocal flags?

tests/Unit/TracingBuilderTest.php
```diff
@@ -85,4 +86,58 @@ private function randomBool()
...
+            ->build();
+        $tracer = $tracing->getTracer();
+
+        $span = $tracer->newTrace();
+        $span->setName('test');
+        $span->start();
+        $span->finish();
+
+        $tracer->flush();
+        $spans = $reporter->flush();
```
adriancole, line 113: I don't fully understand why we are using another word. Ex alwaysReportSpans would make sense to have reporter report. alwaysEmitSpans would makes sense to have emitter emit :)

here's a related post on naming the TL;DR; being that we make enough words at this point https://publicobject.com/2020/06/06/synonyms-are-bad/

drolando: oh yeah that's a better name

drolando:
@jcchavezs any comment on this? Does it look good?

jcchavezs:
Sorry for the delay @drolando. It looks good to me, only one note:

`isSampled` in the recording span would always be Boolean by the time it is
reported. No need to be nullable.

Also, do we need some docs or notes around what alwayaRecordSpans would
cause? Like I know Tracing builder does not have docs for methods but maybe
this is a good one to start with?

Other than that I like it!

src/Zipkin/Recording/Span.php
```diff
@@ -156,7 +163,8 @@ public static function createFromContext(TraceContext $context, Endpoint $localE
             $context->getSpanId(),
             $context->isDebug(),
             $context->isShared(),
-            $localEndpoint
+            $localEndpoint,
+            $context->isSampled()
```
adriancole, line 167: ps one slight thing here is that in brave the "handler" has 2 arguments. since it has the context as one of the args, we don't need to also copy isSampled to the span.

```
  @Override public boolean end(TraceContext context, MutableSpan span, Cause cause) {
    if (!alwaysReportSpans && !Boolean.TRUE.equals(context.sampled())) return true;
--snip--
```

adriancole: I'm not saying to change this, just mentioning a reference. it is fine to also copy it to the span, just it should be immutable (appears to be the case!)

drolando:
> `isSampled` in the recording span would always be Boolean by the time it is reported. No need to be nullable.

This isn't always true, at least in tests. If I create the context with `TraceContext::createAsRoot(DefaultSamplingFlags::createAsEmpty());` then isSampled is `null`.

jcchavezs:
> This isn't always true, at least in tests. If I create the context with TraceContext::createAsRoot(DefaultSamplingFlags::createAsEmpty()); then isSampled is null.

Yes but when making it a span then the sampling decision happens. Anyways I am going to merge this and then investigate that and rebase onto https://github.com/openzipkin/zipkin-php/pull/163

src/Zipkin/Recording/Span.php
```diff
@@ -127,20 +127,27 @@ final class Span
...
+
     private function __construct(
         string $traceId,
         ?string $parentId,
         string $spanId,
         bool $debug,
         bool $shared,
-        Endpoint $localEndpoint
+        Endpoint $localEndpoint,
+        ?bool $isSampled = false
```
jcchavezs, line 142: I will also move this before endpoint as we changed the order already.

tests/Unit/TracingBuilderTest.php
```diff
@@ -85,4 +86,58 @@ private function randomBool()
     {
         return (bool) mt_rand(0, 1);
     }
+
+    public function testAlwaysEmitSpans()
```
jcchavezs, line 90: I will also change this test into `alwaysReportSpans`.

jcchavezs:
Thanks for the great work @drolando !

adriancole:
on the sampled null part. Indeed currently in brave a span must have a sampling decision, but since the context is independent of the span, and it could nave a null decision, then we have to allow it null. At some point we wanted to have "incognito mode" or "pass through" where you don't interfere with anything in the context, but let it be passed across. Especially in that case it could pass around null , but even now it is possible even if not normal.

src/Zipkin/Recording/Span.php
```diff
@@ -225,6 +233,14 @@ public function setRemoteEndpoint(Endpoint $remoteEndpoint): void
         $this->remoteEndpoint = $remoteEndpoint;
     }
 
+    /**
+     * @return bool
+     */
+    public function isSampled(): bool
+    {
+        return $this->isSampled === true;
```
jcchavezs, line 241: Got a question here @adriancole @drolando: what we are saying here is whether this is sampled or not (and then dispatch the span to different reporters accordingly) but what about debug? if something is sampled=false but debug=true then we will dispatch it to the local sampled reporter, isn't?

jcchavezs: Probably it is better we change this into `shouldRecord` which also checks the debug flag?

jcchavezs: Giving a second thought, I think it is responsability of the user to decide what to do when sampled=false and debug=true.

adriancole:
sampled explicitly false and debug true is a nonsense value..the only
nonsense choice  so I wouldnt put much thought into it. depends on the
order set. for example we set sampled when debug is set (debug is
effectively a boosted sample decision)


here's a similar chat about it being a nonsense value
https://github.com/openzipkin/b3-propagation/issues/31#issuecomment-626078005

PR #159, opened 2020-06-16
tests(join-spans): adds tests for shared value in joinSpan.

jcchavezs:
Ping @drolando

### DIFF
--- a/src/Zipkin/Recording/Span.php
+++ b/src/Zipkin/Recording/Span.php
@@ -255,9 +255,12 @@ final class Span
             'parentId' => $this->parentId ? $this->parentId : null,
             'timestamp' => $this->timestamp,
             'duration' => $this->duration,
-            'debug' => $this->debug,
             'localEndpoint' => $this->localEndpoint->toArray(),
         ];
+
+        if ($this->debug === true) {
+            $spanAsArray['debug'] = true;
+        }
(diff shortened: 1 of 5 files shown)


src/Zipkin/Recording/Span.php
```diff
@@ -255,10 +255,13 @@ public function toArray(): array
             'parentId' => $this->parentId ? $this->parentId : null,
             'timestamp' => $this->timestamp,
             'duration' => $this->duration,
-            'debug' => $this->debug,
             'localEndpoint' => $this->localEndpoint->toArray(),
         ];
 
+        if ($this->debug === true) {
```
jcchavezs, line 261: unrelated but this change is good.

tests/Unit/TracerTest.php
```diff
@@ -439,11 +413,47 @@ public function __invoke(int $a, int $b)
...
-        $tracer->flush();
-        $spans = $reporter->flush();
+        $spans = $flusher();
         $this->assertCount(1, $spans);
 
         $span = $spans[0]->toArray();
         $this->assertEquals('too small values', $span['tags']['error']);
     }
+
+    public function testJoinSpans()
```
jcchavezs, line 423: This is the important change @drolando

PR #160, opened 2020-06-18
feat(sampling-flags): adds withSampled method for SamplingFlags.

jcchavezs:
Currently there it is only possible to override the sampling decision for trace context but not for sampling flags. This formalizes the use case to override the sampling in the SamplingFlags interface.

The use case is the popular "requestSampler" which changes the sampler decision based on a request (both for client and server).

The use case is in the usage of `Tracer::nextSpanWithSampler` as used in https://github.com/jcchavezs/zipkin-instrumentation-symfony/pull/70/files#diff-46b82fded07ac22b103293cbc8c3ebbeR60-R63

Ping @adriancole

### DIFF
--- a/src/Zipkin/Propagation/DefaultSamplingFlags.php
+++ b/src/Zipkin/Propagation/DefaultSamplingFlags.php
@@ -22,59 +22,36 @@ final class DefaultSamplingFlags implements SamplingFlags
         $this->isDebug = $isDebug;
     }
 
-    /**
-     * @param bool|null $isSampled
-     * @param bool $isDebug
-     * @return DefaultSamplingFlags
-     */
     public static function create(?bool $isSampled, bool $isDebug = false): self
     {
         return new self($isSampled, $isDebug);
     }
 
-    /**
-     * @return DefaultSamplingFlags
-     */
     public static function createAsEmpty(): self
     {
         return new self(SamplingFlags::EMPTY_SAMPLED, SamplingFlags::EMPTY_DEBUG);
     }
 
-    /**
-     * @return DefaultSamplingFlags
-     */
     public static function createAsSampled(): self
     {
         return new self(true, false);
     }
 
-    /**
-     * @return DefaultSamplingFlags
-     */
     public static function createAsNotSampled(): self
     {
         return new self(false, false);
     }
 
-    /**
-     * @return DefaultSamplingFlags
-     */
     public static function createAsDebug(): self
     {
         return new self(null, true);
     }
 
-    /**
-     * @return bool|null
-     */
     public function isSampled(): ?bool
     {
         return $this->isSampled;
     }
 
-    /**
-     * @return bool
-     */
     public function isDebug(): bool
     {
         return $this->isDebug;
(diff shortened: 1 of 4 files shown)


src/Zipkin/Propagation/TraceContext.php
```diff
@@ -248,17 +244,11 @@ public function isEmpty(): bool
...
-     */
     public function isEqual(SamplingFlags $samplingFlags): bool
     {
         return ($samplingFlags instanceof TraceContext)
             && $this->traceId === $samplingFlags->traceId
             && $this->spanId === $samplingFlags->spanId
-            && $this->parentId === $samplingFlags->parentId
-            && $this->isSampled === $samplingFlags->isSampled
-            && $this->isDebug === $samplingFlags->isDebug;
+            && $this->parentId === $samplingFlags->parentId;
```
jcchavezs, line 252: Removing the `isSampled` and `isDebug` from the comparison as even if sampling decision changed the context remains the same.

src/Zipkin/Propagation/DefaultSamplingFlags.php
```diff
@@ -86,13 +63,14 @@ public function isEmpty(): bool
...
-     */
     public function isEqual(SamplingFlags $samplingFlags): bool
     {
-        return $this->isDebug() === $samplingFlags->isDebug()
-            && $this->isSampled() === $samplingFlags->isSampled();
+        return $this->isDebug === $samplingFlags->isDebug
+            && $this->isSampled === $samplingFlags->isSampled;
+    }
+
+    public function withSampled(bool $isSampled): DefaultSamplingFlags
```
jcchavezs, line 72: This is the important change.

src/Zipkin/Propagation/SamplingFlags.php
```diff
@@ -9,21 +9,13 @@ interface SamplingFlags
...
 
     public function isEmpty(): bool;
 
-    /**
-     * @param SamplingFlags $samplingFlags
-     * @return bool
-     */
     public function isEqual(SamplingFlags $samplingFlags): bool;
+
+    public function withSampled(bool $isSampled): SamplingFlags;
```
jcchavezs, line 20: This is important too.

src/Zipkin/Propagation/DefaultSamplingFlags.php
adriancole: maybe wait until you use it? or use it?

I feel it is best to add the feature with what uses it, or worst case a code doc and test.

ex here's the use in brave's http server handler:
```java
    Boolean sampled = extracted.sampled();
    // only recreate the context if the http sampler made a decision
    if (sampled == null && (sampled = sampler.trySample(request)) != null) {
      extracted = extracted.sampled(sampled.booleanValue());
    }
```
https://github.com/openzipkin/brave/blob/master/instrumentation/http/src/main/java/brave/http/HttpServerHandler.java#L148-L152

jcchavezs: Sorry I should have linked the use case. We need something like Tracer::nextSpanWithSampler https://github.com/openzipkin/zipkin-php/pull/160/files#diff-ced4309e6e2dfd1e487df57084f07566R252 to be used in https://github.com/jcchavezs/zipkin-instrumentation-symfony/pull/70/files#diff-46b82fded07ac22b103293cbc8c3ebbeR60-R63

jcchavezs: Added another example for a different instrumentation: https://github.com/openzipkin/zipkin-php/pull/149/files#diff-aa2738d45e9308a2c88d57d7e7dde018R54-R59

adriancole: sorry I guess what I meant was that something using the code should live in this repo. If it is for code in another repo, maybe make a small copy of the call site in the test tree with a link to where its used. This can prevent things from accidentally drifting (even if you are on top of things at the moment)

jcchavezs: As mentioned internally, this is the code living in this repo: https://github.com/openzipkin/zipkin-php/pull/160/files#diff-ced4309e6e2dfd1e487df57084f07566R252

PR #162, opened 2020-06-24
feat(http-server): adds middleware instrumentation.

jcchavezs:
This PR adds support for instrumentation in PSR middlewares.

Ping @adriancole


src/Zipkin/Instrumentation/Http/Server/DefaultParser.php
```diff
@@ -0,0 +1,39 @@
...
+
+    public function response(ResponseInterface $response, TraceContext $context, SpanCustomizer $span): void
+    {
+        $span->tag(Tags\HTTP_STATUS_CODE, (string) $response->getStatusCode());
+        if ($response->getStatusCode() > 399) {
+            $span->tag(Tags\ERROR, (string) $response->getStatusCode());
+        }
+    }
+
+    public function error(Throwable $e, TraceContext $context, SpanCustomizer $span): void
```
adriancole, line 35: fyi in brave we don't do this at the abstraction of http etc as often the throwable is generically parsable. Also in case it goes to another system, we don't set it at all (as it could be used in another field called exception for example)
ex we opt out like this:
```java
      if (error != null) return; // the call site used Span.error
```
https://github.com/openzipkin/brave/blob/master/instrumentation/http/src/main/java/brave/http/HttpResponseParser.java#L115

And defer to zipkin handler to use an error parser which internally does this:
```java
  static String parse(Throwable error) {
    if (error == null) throw new NullPointerException("error == null");
    String message = error.getMessage();
    if (message != null) return message;
    if (error.getClass().isAnonymousClass()) { // avoids ""
      return error.getClass().getSuperclass().getSimpleName();
    }
    return error.getClass().getSimpleName();
  }
```

In the zipkin variant of the error reporter we generically convent errors using that parser:

https://github.com/openzipkin/zipkin-reporter-java/blob/master/brave/src/main/java/zipkin2/reporter/brave/ConvertingSpanReporter.java#L89-L93

In other words we used to put `$span->tag(Tags\ERROR, $e->getMessage())` in different code, but then had to later centralize it, and then later defer it as when you use other tools like x-ray the tag shape is different.

src/Zipkin/Instrumentation/Http/Server/Middleware.php
```diff
@@ -0,0 +1,87 @@
...
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $extractedContext = ($this->extractor)($request);
+
+        if ($extractedContext instanceof TraceContext) {
+            $span = $this->tracer->joinSpan($extractedContext);
+        } elseif ($this->requestSampler === null) {
+            $span = $this->tracer->nextSpan($extractedContext);
+        } else {
+            $span = $this->tracer->nextSpanWithSampler(
```
adriancole, line 56: In brave, we do this logic differently as it keeps the tracer api small which is helpful as sometimes there are many "if statements" inside, and keeping all inside nextSpan(extracted) helps sanity a little.

```java
  Span nextSpan(TraceContextOrSamplingFlags extracted, HttpServerRequest request) {
    Boolean sampled = extracted.sampled();
    // only recreate the context if the http sampler made a decision
    if (sampled == null && (sampled = sampler.trySample(request)) != null) {
      extracted = extracted.sampled(sampled.booleanValue());
    }
    return extracted.context() != null
      ? tracer.joinSpan(extracted.context())
      : tracer.nextSpan(extracted);
  }
```

src/Zipkin/Instrumentation/Http/Server/Middleware.php
```diff
@@ -0,0 +1,87 @@
...
+                $this->requestSampler,
+                [$request],
+                $extractedContext
+            );
+        }
+        $spanCustomizer = null;
```
adriancole, line 62: instead of all the null checking, why not check if noop and early return `$handler->handle($request);`?

src/Zipkin/Instrumentation/Http/Server/Parser.php
```diff
@@ -0,0 +1,53 @@
...
+use Throwable;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Parser includes the methods to obtain meaningful span information
+ * out of HTTP request/response elements.
+ */
+interface Parser
```
adriancole, line 18: design note:

we used to have a design like this, but we don't any more. I already mentioned about the error handling in an earlier comment.

instead we have lambda compatible (single method) functions for the request and response side.

HttpRequestParser -> `void parse(HttpRequest request, TraceContext context, SpanCustomizer span)`
HttpResponseParser -> `void parse(HttpResponse response, TraceContext context, SpanCustomizer span)`

At least in Java, this allows easier composition. Ex you can rely on the default for span name etc and add a tag like so

```java
    .clientRequestParser((req, context, span) -> {
      HttpClientRequestParser.DEFAULT.parse(req, context, span);
      HttpTags.URL.tag(req, context, span); // add the url in addition to defaults
    })
```

Most people need to change request not response tags anyway so it has little impact to split them, though the rationale may not be the same here.

src/Zipkin/Instrumentation/Http/Server/RequestHandler.php
```diff
@@ -0,0 +1,94 @@
...
+use Zipkin\SpanCustomizerShield;
+use Zipkin\Propagation\TraceContext;
+use Zipkin\Propagation\RequestHeaders;
+use Zipkin\Kind;
+use Throwable;
+use Psr\Http\Server\RequestHandlerInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+final class RequestHandler implements RequestHandlerInterface
```
adriancole, line 17: I'm surprised to see almost identical logic here vs middleware.. In brave the main idea of creating the handlers are so that you can re-use them in different types of call sites.. I'll show an example of RPC handler as it doesn't have the clutter our http one does as it has no deprecated code in it, but does the same as what brave's http does https://github.com/openzipkin/brave/blob/master/instrumentation/rpc/src/main/java/brave/rpc/RpcServerHandler.java

src/Zipkin/Instrumentation/Http/Server/DefaultParser.php
jcchavezs: So if I understand correctly you don't treat error as something being part of an http tracing but instead you have an error parser that transform the error into something customized? I see that being feasible because Tags are objects in brave, in zipkin-php they are strings (I remember I wanted them to be stringable objects first but objects in PHP are not as cheap as in Java). To be honest, we are still in time to do such change (accept anything for tags and stringify them afterwards) because we are just introducing stronger typing and even more, if we don't do that change we won't be able to do it in the future but I am not entirely sure of the implications of such a change.

adriancole: the jist of the change is you store the Throwable and only parse it into a String tag at the last moment (converting to json for example). So our field used is `MutableSpan.error` which is separate from `MutableSpan.tags`. 

here's the notes about that field which I think says all there is to be said (at least that I know :)) https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/handler/MutableSpan.java#L49-L86

jcchavezs: I see. I did not know `Span.error` was a thing in brave. Let me dig into it.

src/Zipkin/Instrumentation/Http/Server/Middleware.php
jcchavezs: Totally. Let me change it.

src/Zipkin/Instrumentation/Http/Server/DefaultParser.php
jcchavezs: Done: https://github.com/openzipkin/zipkin-php/pull/162/files#diff-7f86d716dfda08180da2ce29601ff41bR83

src/Zipkin/Instrumentation/Http/Server/RequestHandler.php
jcchavezs: I removed the RequestHandler. For instrumentation we will use middlewares.

src/Zipkin/Instrumentation/Http/Server/Middleware.php
jcchavezs: Check https://github.com/openzipkin/zipkin-php/pull/162/commits/eb5bee4d690deaf9e482b5bec9beb58ea062ff16 where I added the `nextSpanHandler` into the `ServerTracing`.

jcchavezs:
@adriancole PTAL

jcchavezs:
@adriancole I addressed most of your comments. I turned the requestSampler into the `nextSpanHandler` in the serverTracing and I am pretty happy about it. If this ends up being the final design, I will replicate it in my symfony instrumentation.

src/Zipkin/Instrumentation/Http/Server/README.md
```diff
@@ -0,0 +1,41 @@
...
+composer require psr/http-server-middleware
+```
+
+## Usage
+
+In this example we use [fast-route](https://github.com/middlewares/fast-route) and [request-handler](https://github.com/middlewares/request-handler) middlewares but any HTTP server middleware supporting PSR15 middlewares will work.
+
+```php
+use Zipkin\Instrumentation\Http\Server\Middleware as ZipkinMiddleware;
+use Zipkin\Instrumentation\Http\Server\ServerTracing;
```
adriancole, line 19: I see this is to avoid clash with normal Tracing. since RpcTracing may come next and that would have a  client and server side.. and be configured possibly in the same file.. does it make more sense to use HttpTracing instead?

src/Zipkin/Instrumentation/Http/Server/ServerTracing.php
```diff
@@ -0,0 +1,88 @@
...
+    {
+        return $this->tracing;
+    }
+
+    public function getParser(): Parser
+    {
+        return $this->parser;
+    }
+
+    public function getNextSpanHandler(): callable
```
adriancole, line 65: if this is only used once maybe it can live at its call site? having it here requires you to say why and how it is used

src/Zipkin/Instrumentation/Http/Client/HttpClientTracing.php
```diff
@@ -11,7 +11,7 @@
  * ClientTracing includes all the elements needed to instrument a
  * HTTP client.
  */
-class ClientTracing
+class HttpClientTracing
```
jcchavezs, line 14: Changed the name @adriancole

src/Zipkin/Instrumentation/Http/Server/Psr15/Middleware.php
```diff
@@ -0,0 +1,92 @@
...
+        $this->extractor = $tracing->getTracing()->getPropagation()->getExtractor(new RequestHeaders());
+        $this->parser = $tracing->getParser();
+        $this->requestSampler = $tracing->getRequestSampler();
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $extractedContext = ($this->extractor)($request);
+
+        if ($extractedContext instanceof TraceContext) {
```
jcchavezs, line 53: For now I am going to move this into a private method with a meaningful name.

src/Zipkin/Instrumentation/Http/Server/Psr15/Middleware.php
```diff
@@ -0,0 +1,92 @@
...
+use Zipkin\Kind;
+use Zipkin\Instrumentation\Http\Server\Parser;
+use Zipkin\Instrumentation\Http\Server\HttpServerTracing;
+use Throwable;
+use Psr\Http\Server\RequestHandlerInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+final class Middleware implements MiddlewareInterface
```
jcchavezs, line 19: This is the PSR15 implementation using the server tracing.

src/Zipkin/Instrumentation/Http/Client/HttpClientTracing.php
```diff
@@ -25,23 +25,19 @@ class ClientTracing
...
+     * request.
      *
      * @var callable(RequestInterface):?bool|null
      */
     private $requestSampler;
 
     public function __construct(
         Tracing $tracing,
-        Parser $parser = null,
+        Parser $parser,
```
jcchavezs, line 36: Removing the null as it was intended for the HttpClientTracing to use a default one but since the default one requires a request type to be defined we require it as a parameter.

src/Zipkin/Instrumentation/Http/Server/Parser.php
jcchavezs: I'd keep them together as IMHO it is easier to rationale about them together and also because I don't want to create too much classes around this.

jcchavezs:
@adriancole please take another look. I addressed most of your comments, I postponed the `nextSpan` thing and started this PR https://github.com/openzipkin/zipkin-php/pull/169 for starting the laravel instrumentation using this API (additional to the PSR one).

src/Zipkin/Instrumentation/Http/Server/Psr15/DefaultParser.php
```diff
@@ -0,0 +1,61 @@
...
+use Zipkin\Instrumentation\Http\Server\Parser;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * DefaultParser contains the basic logic for turning request/response information
+ * into span name and tags. Implementors can use this as a base parser to reduce
+ * boilerplate.
+ */
+class DefaultParser implements Parser
```
adriancole, line 19: this makes me curious. in brave, the parser is intended to be defined in a way that is not library specific. This allows it to be generated from static configuration that doesn't know anything about the types that would be used. What I'm seeing here is that DefaultParser and the data policy is being independently defined in different libraries.

I would recommend instead to keep parser language agnostic. similar to brave, make the caller of HttpTracing localize the Request interface. ex. `Zipkin\Instrumentation\Http\HttpRequest` is the parameter of parser, and `Zipkin\Instrumentation\Http\Server\Psr15\SomethingRequest` which can be package private, implements it.

that different libraries treat data differently is normal, but if the "parser of http requests" requires knowing which library is in use, we lose the ability to change tagging policy independent of library, ex from yaml.

ex notice here that "how" an http header is parsed is not something the user should care about.. it is just an implementation detail.

https://github.com/openzipkin/brave/blob/master/instrumentation/okhttp3/src/main/java/brave/okhttp3/TracingInterceptor.java#L93-L129

doing so allows you to configure ITs at global level, such as I think we also do in zipkin-js etc
https://github.com/openzipkin/brave/blob/master/instrumentation/http-tests/src/main/java/brave/test/http/ITHttpClient.java#L234-L237

that allows configuration or global config to change the sampling policy. Ex if X-Ray needs url you add code for that without caring knowing or linking to library is PSR etc... to do that.. make sense?

jcchavezs:
@adriancole I just added the Request/Response types and aligned the parsers. PTAL.

src/Zipkin/Instrumentation/Http/Server/DefaultParser.php
```diff
@@ -0,0 +1,41 @@
...
+
+    public function request(Request $request, TraceContext $context, SpanCustomizer $span): void
+    {
+        $span->tag(Tags\HTTP_METHOD, $request->getMethod());
+        $span->tag(Tags\HTTP_PATH, $request->getPath() ?: '/');
+    }
+
+    public function response(Response $response, TraceContext $context, SpanCustomizer $span): void
+    {
+        $span->tag(Tags\HTTP_STATUS_CODE, (string) $response->getStatusCode());
```
adriancole, line 32: in brave I think we managed to have the parsers the same client and server.. this can help with some copy/paste stuff around tag interpretation. Ex in brave the parser is HttpResponseParser not HttpServerResponseParser as we don't have anything different by default client vs server side.

src/Zipkin/Instrumentation/Http/Server/Parser.php
```diff
@@ -0,0 +1,42 @@
...
+{
+    /**
+     * spanName returns an appropiate span name based on the request,
+     * usually the HTTP method is enough (e.g GET or POST) but ideally
+     * the http.route is desired (e.g. /user/{user_id}).
+     */
+    public function spanName(Request $request): string;
```
adriancole, line 21: similar comment, that the client and server side have two different types that do the same things, means there are some opportunity for copy/paste problems.

src/Zipkin/Instrumentation/Http/Server/Psr15/README.md
```diff
@@ -0,0 +1,42 @@
...
+$fastRouteDispatcher = FastRoute\simpleDispatcher(function (FastRoute\RouteCollector $r) {
+    $r->get('/hello/{name}', HelloWorldController::class);
+});
+
+// Creates tracing component
+$tracing = TracingBuilder::create()
+            ->havingLocalServiceName('my_service')
+            ->build();
+
+$httpClientTracing = new ServerTracing($tracing, new DefaultParser);
```
adriancole, line 32: again user shouldn't have to instantiate a default parser anymore

src/Zipkin/Instrumentation/Http/Server/Request.php
```diff
@@ -0,0 +1,24 @@
...
+     * is returned. Null indicates this instrumentation doesn't understand http routes.
+     *
+     * <p>The route is associated with the request, but it may not be visible until response
+     * processing. The reasons is that many server implementations process the request before they can
+     * identify the route. Parsing should expect this and look at {@link HttpResponse#route()} as
+     * needed.
+     *
+     * @see Tags\HTTP_ROUTE
+     */
+    public function getRoute(): ?string;
```
adriancole, line 23: route is also possible on some clients, not sure in PHP. It makes things easier to assume everything can see it and just handle if it is not visible vs 2 different parser hierarchies.

src/Zipkin/Instrumentation/Http/Server/Response.php
```diff
@@ -0,0 +1,24 @@
...
+     * is returned. Null indicates this instrumentation doesn't understand http routes.
+     *
+     * <p>The route is associated with the request, but it may not be visible until response
+     * processing. The reasons is that many server implementations process the request before they can
+     * identify the route. Parsing should expect this and look at {@link HttpResponse#route()} as
+     * needed.
+     *
+     * @see Tags\HTTP_ROUTE
+     */
+    public function getRoute(): ?string;
```
adriancole, line 23: technically eventhough this is visible in the response, it is a part of request data. once we added response.request() we were able to keep this data in the more logical place https://github.com/openzipkin/brave/blob/master/instrumentation/http/src/main/java/brave/http/HttpRequest.java#L92-L106

tests/Unit/Instrumentation/Http/Client/Psr18/ClientTest.php
```diff
@@ -30,7 +31,7 @@ private static function createTracing(): array
         $tracer = $tracing->getTracer();
 
         return [
-            new ClientTracing($tracing),
+            new HttpClientTracing($tracing, new DefaultParser),
```
adriancole, line 34: same comment

src/Zipkin/Instrumentation/Http/Client/DefaultParser.php
```diff
@@ -7,28 +7,34 @@
...
 
+/**
+ * DefaultParser contains the basic logic for turning request/response information
+ * into span name and tags. Implementors can use this as a base parser to reduce
+ * boilerplate.
+ */
 class DefaultParser implements Parser
 {
-    public function spanName(RequestInterface $request): string
+    public function spanName(Request $request): string
```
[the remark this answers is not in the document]
jcchavezs: Makes sense. Actually I was about to change this as using the API felt weird.

src/Zipkin/Instrumentation/Http/Client/DefaultParser.php
```diff
@@ -7,28 +7,34 @@
...
     }
 
-    public function response(ResponseInterface $response, TraceContext $context, SpanCustomizer $span): void
+    public function response(Response $response, TraceContext $context, SpanCustomizer $span): void
     {
-        $span->tag(Tags\HTTP_STATUS_CODE, (string) $response->getStatusCode());
-        if ($response->getStatusCode() > 399) {
-            $span->tag(Tags\ERROR, (string) $response->getStatusCode());
+        $statusCode = $response->getStatusCode();
+        $span->tag(Tags\HTTP_STATUS_CODE, (string) $statusCode);
```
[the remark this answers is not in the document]
jcchavezs: Good reminder. I wanted to introduce this before but I wasn't sure this applied for both client and server, by that time I wouldn't dive into the http instrumentation design so many doubts and I did not want to introduce a change I did not understand. Anyways good time to introduce it.

src/Zipkin/Instrumentation/Http/Client/Psr18/README.md
```diff
@@ -17,14 +17,15 @@ In this example we use Guzzle 7 but any HTTP client supporting PSR18 clients wil
...
+use Zipkin\Instrumentation\Http\Client\HttpClientTracing;
+use Zipkin\Instrumentation\Http\Client\Psr\Client as ZipkinClient;
+use Zipkin\Instrumentation\Http\Client\Psr\DefaultParser;
 
 $tracing = TracingBuilder::create()
             ->havingLocalServiceName('my_service')
             ->build();
 
-$httpClientTracing = new ClientTracing($tracing);
+$httpClientTracing = new HttpClientTracing($tracing, new DefaultParser);
```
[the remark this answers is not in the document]
jcchavezs: Yeah, totally. Also when using default parser we don't need to pass it. Honestly having to pass the `DefaultParser` per library was akward.

src/Zipkin/Instrumentation/Http/Server/Request.php
jcchavezs: I don't think it is possible to pass router in PHP clients, I haven't seen any example. Since we know use abstract classes I think it is OK to not to add route yet.

src/Zipkin/Instrumentation/Http/Server/Response.php
jcchavezs: My concern about that is that if route is not there yet at request parsing, having the ability to parse later means we mutated the request. In any case I will check first what happens in laravel or symfony before settling on this.

jcchavezs: In the end I removed it as now we use abstract classes. I will readd it when the time comes.

jcchavezs:
@adriancole I think I addressed all the feedback so please take another look and feel free to merge if you agree with all.

PR #163, opened 2020-07-02
feat(span-error): adds support for span error.

jcchavezs:
This PR adds support for span.error.

Ping @adriancole

### DIFF
--- a/src/Zipkin/DefaultErrorParser.php
+++ b/src/Zipkin/DefaultErrorParser.php
@@ -11,9 +11,4 @@ class DefaultErrorParser implements ErrorParser
     {
         return [Tags\ERROR => $e->getMessage()];
     }
-
-    public function parseAnnotations(Throwable $e): array
-    {
-        return [];
-    }
 }
--- a/src/Zipkin/DefaultErrorParser.php
+++ b/src/Zipkin/DefaultErrorParser.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Zipkin;
+
+use Zipkin\Tags;
+use Throwable;
+
+class DefaultErrorParser implements ErrorParser
+{
+    public function parseTags(Throwable $e): array
+    {
+        return [Tags\ERROR => $e->getMessage()];
+    }
+
+    public function parseAnnotations(Throwable $e): array
+    {
+        return [];
+    }
+}
--- a/src/Zipkin/Recording/ReadbackSpan.php
+++ b/src/Zipkin/Recording/ReadbackSpan.php
@@ -9,6 +9,10 @@ use Zipkin\ErrorParser;
 use Zipkin\Endpoint;
 use Throwable;
 
+/**
+ * ReadbackSpan is an interface for accessing the recording
+ * span without the possibility to mutate it.
+ */
 interface ReadbackSpan
 {
     public function getSpanId(): string;
--- a/src/Zipkin/Recording/ReadbackSpan.php
+++ b/src/Zipkin/Recording/ReadbackSpan.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Zipkin\Recording;
+
+use Zipkin\Reporters\SpanSerializer;
+use Zipkin\ErrorParser;
+use Zipkin\Endpoint;
+use Throwable;
+
+interface ReadbackSpan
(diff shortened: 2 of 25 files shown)


src/Zipkin/Recording/ReadbackSpan.php
```diff
@@ -0,0 +1,25 @@
...
+declare(strict_types=1);
+
+namespace Zipkin\Recording;
+
+use Zipkin\Reporters\SpanSerializer;
+use Zipkin\ErrorParser;
+use Zipkin\Endpoint;
+use Throwable;
+
+interface ReadbackSpan
```
jcchavezs, line 12: This interface allows the serializer to access the MutableSpan but without having the possibility to change anything.

src/Zipkin/Tracer.php
```diff
@@ -329,33 +328,38 @@ public function inSpan(
...
         try {
             $result = \call_user_func_array($fn, $args);
-            ($resultParser)($spanCustomizer, $result, null);
+            if ($resultParser !== null) {
+                ($resultParser)($result, $span->getContext(), $spanCustomizer);
+            }
             return $result;
         } catch (Throwable $e) {
-            ($resultParser)($spanCustomizer, null, $e);
+            $span->setError($e);
```
jcchavezs, line 354: Notice here we use the `span::setError` as it will be used in instrumentations.

src/Zipkin/Reporters/JsonV2Serializer.php
```diff
@@ -0,0 +1,67 @@
...
+use Zipkin\Recording\ReadbackSpan;
+use Zipkin\DefaultErrorParser;
+
+class JsonV2Serializer implements SpanSerializer
+{
+    private $errorParser;
+
+    public function __construct(ErrorParser $errorParser = null)
+    {
+        $this->errorParser = $errorParser ?? new DefaultErrorParser();
```
jcchavezs, line 16: I delegated the errorParser responsibility into the serializer, mostly to avoid incosistencies where the `RealSpan` has the error and the error tag and same for MutableSpan, hence the tagging happens when the serialization.

src/Zipkin/DefaultErrorParser.php
```diff
@@ -0,0 +1,19 @@
...
+use Throwable;
+
+class DefaultErrorParser implements ErrorParser
+{
+    public function parseTags(Throwable $e): array
+    {
+        return [Tags\ERROR => $e->getMessage()];
+    }
+
+    public function parseAnnotations(Throwable $e): array
```
adriancole, line 15: IMHO I wouldn't recommend adding this, as annotations are not well supported. Focusing on ErrorTagParser will make it super clear (especially that this is a single tag value)

src/Zipkin/Recording/ReadbackSpan.php
```diff
@@ -0,0 +1,25 @@
...
+{
+    public function getSpanId(): string;
+    public function getTraceId(): string;
```
adriancole, line 15: hmm wouldn't it need parentId, shared and debug?

src/Zipkin/Reporters/JsonV2Serializer.php
```diff
@@ -0,0 +1,67 @@
...
+            if ($span->getRemoteEndpoint() !== null) {
+                $spanAsArray['remoteEndpoint'] = $span->getRemoteEndpoint()->toArray();
+            }
+    
+            if (!empty($span->getAnnotations())) {
+                $spanAsArray['annotations'] = $span->getAnnotations() + $this->errorParser->parseAnnotations($span->getError());
+            }
+    
+            if (!empty($span->getTags())) {
+                $spanAsArray['tags'] = $span->getTags() + $this->errorParser->parseTags($span->getError());
```
adriancole, line 59: make sure you have a test where an existing "error" tag isn't clobbered. This is one good reason to have errorTagParser as you know what the rules should be if it can only add "error" tag

jcchavezs: There you go https://github.com/openzipkin/zipkin-php/pull/163/commits/661f9958cf364f7d904e6c66f0988599b2a795a7#diff-da7d21b7a5a702c26a9d8ffce6d9cb1cR41

PR #164, opened 2020-07-05
chore: adds tweaks to Span::isSampled and TracerTest.

jcchavezs:


tests/Unit/TracerTest.php
```diff
@@ -214,6 +214,7 @@ public function testNextSpanIsCreatedFromSamplingFlags($sampled, $debug)
         $span = $tracer->nextSpan($samplingFlags);
 
         $this->assertSameSamplingFlags($samplingFlags, $span->getContext());
+        $this->assertNotNull($span->getContext()->isSampled());
```
jcchavezs, line 217: This makes sure when there is not sampling decision, we have a value anyways.

PR #165, opened 2020-07-06
fix(propagation): fixes logic around sampling and debug to improve the API usability.

jcchavezs:
From now on, debug=true will force a sampled=true value in the TraceContext, making it easier for users to just check $context->isSampled() instead of also checking the $context->isDebug().

Ping @adriancole

### DIFF
--- a/src/Zipkin/Propagation/B3.php
+++ b/src/Zipkin/Propagation/B3.php
@@ -4,13 +4,13 @@ declare(strict_types=1);
 
 namespace Zipkin\Propagation;
 
-use InvalidArgumentException;
+use Zipkin\Propagation\TraceContext;
+use Zipkin\Propagation\RemoteSetter;
+use Zipkin\Propagation\Exceptions\InvalidTraceContextArgument;
 use Zipkin\Kind;
 use Psr\Log\NullLogger;
 use Psr\Log\LoggerInterface;
-use Zipkin\Propagation\RemoteSetter;
-use Zipkin\Propagation\TraceContext;
-use Zipkin\Propagation\Exceptions\InvalidTraceContextArgument;
+use InvalidArgumentException;
 
 /**
  * @see https://github.com/openzipkin/b3-propagation
@@ -251,12 +251,12 @@ final class B3 implements Propagation
             }
         }
 
-        if ($traceContext->isSampled() !== null) {
-            $setter->put($carrier, self::SAMPLED_NAME, $traceContext->isSampled() ? '1' : '0');
-        }
-
         if ($traceContext->isDebug()) {
             $setter->put($carrier, self::FLAGS_NAME, '1');
+        } elseif ($traceContext->isSampled() !== null) {
+            // according to https://github.com/openzipkin/b3-propagation#debug-flag if debug
+            // is set we don't pass the sampling bit.
+            $setter->put($carrier, self::SAMPLED_NAME, $traceContext->isSampled() ? '1' : '0');
         }
     }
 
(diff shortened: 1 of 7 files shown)


src/Zipkin/Propagation/B3.php
```diff
@@ -4,13 +4,13 @@
 
 namespace Zipkin\Propagation;
 
-use InvalidArgumentException;
+use Zipkin\Propagation\TraceContext;
```
jcchavezs, line 7: Not important, just sorting.

src/Zipkin/Propagation/B3.php
```diff
@@ -251,12 +251,12 @@ private static function injectMultiValues(Setter $setter, TraceContext $traceCon
             }
         }
 
-        if ($traceContext->isSampled() !== null) {
-            $setter->put($carrier, self::SAMPLED_NAME, $traceContext->isSampled() ? '1' : '0');
-        }
-
         if ($traceContext->isDebug()) {
             $setter->put($carrier, self::FLAGS_NAME, '1');
+        } elseif ($traceContext->isSampled() !== null) {
```
jcchavezs, line 256: This piece makes it possible to only pass the debug header if debug=true and skip the sampled.

src/Zipkin/Propagation/B3.php
```diff
@@ -364,20 +364,23 @@ public static function parseSingleValue(string $value): ?SamplingFlags
...
+        $isDebugRaw = $getter->get($carrier, self::FLAGS_NAME);
+        $isDebug = $isDebugRaw === '1';
 
         $isSampled = SamplingFlags::EMPTY_SAMPLED;
-        if ($isSampledRaw !== null) {
-            if ($isSampledRaw === '1' || \strtolower($isSampledRaw) === 'true') {
-                $isSampled = true;
-            } elseif ($isSampledRaw === '0' || \strtolower($isSampledRaw) === 'false') {
-                $isSampled = false;
+        if ($isDebug) {
```
jcchavezs, line 371: If debug=true we don't bother looking at sampled when extracting headers.

src/Zipkin/Propagation/DefaultSamplingFlags.php
```diff
@@ -24,7 +24,7 @@ private function __construct(?bool $isSampled, bool $isDebug)
 
     public static function create(?bool $isSampled, bool $isDebug = false): self
     {
-        return new self($isSampled, $isDebug);
+        return new self($isDebug ?: $isSampled, $isDebug);
```
jcchavezs, line 27: if debug=true we pass sampled-true when creating samplingFlags.

src/Zipkin/Propagation/DefaultSamplingFlags.php
```diff
@@ -74,6 +74,6 @@ public function isEqual(SamplingFlags $samplingFlags): bool
      */
     public function withSampled(bool $isSampled): SamplingFlags
     {
-        return new self($isSampled, $this->isDebug);
+        return new self($isSampled, SamplingFlags::EMPTY_DEBUG);
```
jcchavezs, line 77: I think there is no reason to keep the debug=true if sampled is being changed by the user. Any thoughts @adriancole

adriancole: it would be nuanced and worth going with something as unlikely anyone will complain or notice either choice.

jcchavezs: SamplingFlags::EMPTY_DEBUG = false but let's make it explicit.

jcchavezs:
@adriancole PTAL

PR #170, opened 2020-07-29
tests(http): covers http client/server request/response.

jcchavezs:
This PR covers the Request/Response value objects for client and server abstractions. It also makes it open the return type for `Response::getRequest` so the extending classes can decide.


jcchavezs:
Error in appveyor for the PR is related to the download of the image so I am ignoring it.

PR #180, opened 2020-10-19
feat(http): adds support for route in response object for late parsing.

jcchavezs:
This PR adds support for including the route for late parsing, mostly for those frameworks where the route is not available until very late in the request/response lifecycle (e.g. symfony).

e.g. https://github.com/jcchavezs/zipkin-instrumentation-symfony/blob/master/src/ZipkinBundle/KernelListener.php#L218

### DIFF
--- a/src/Zipkin/Instrumentation/Http/Server/DefaultHttpServerParser.php
+++ b/src/Zipkin/Instrumentation/Http/Server/DefaultHttpServerParser.php
@@ -17,11 +17,11 @@ use Zipkin\Instrumentation\Http\Server\HttpServerParser;
 class DefaultHttpServerParser implements HttpServerParser
 {
     /**
-     * spanName returns an appropiate span name based on the request,
+     * spanNameFromRequest returns an appropiate span name based on the request,
      * usually the HTTP method is enough (e.g GET or POST) but ideally
-     * the http.route is desired (e.g. /user/{user_id}).
+     * the verb+route is desired (e.g. GET /user/{user_id}).
      */
-    protected function spanName(Request $request): string
+    protected function spanNameFromRequest(Request $request): string
     {
         return $request->getMethod()
             . ($request->getRoute() === null ? '' : ' ' . $request->getRoute());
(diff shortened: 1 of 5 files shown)


src/Zipkin/Instrumentation/Http/Server/DefaultHttpServerParser.php
```diff
@@ -17,11 +17,11 @@
...
 {
     /**
-     * spanName returns an appropiate span name based on the request,
+     * spanNameFromRequest returns an appropiate span name based on the request,
      * usually the HTTP method is enough (e.g GET or POST) but ideally
-     * the http.route is desired (e.g. /user/{user_id}).
+     * the verb+route is desired (e.g. GET /user/{user_id}).
      */
-    protected function spanName(Request $request): string
+    protected function spanNameFromRequest(Request $request): string
```
jcchavezs, line 24: while this could be a breaking change because this class can be extended, it is very early in this version and I don't think there are implementors yet (zipkin-instrumentation-symfony is the first one and yet it does not extend it).

src/Zipkin/Instrumentation/Http/Server/DefaultHttpServerParser.php
```diff
@@ -32,16 +32,41 @@ protected function spanName(Request $request): string
...
+    {
+        return $request->getMethod()
+            . ($request->getRoute() === null ? '' : ' ' . $request->getRoute());
+    }
+
+    /**
+     * spanNameFromResponse returns an appropiate span name based on the response's request,
+     * usually seeking for a better name than the HTTP method (e.g. GET /user/{user_id}).
+     */
+    protected function spanNameFromResponse(Response $response): ?string
```
adriancole, line 50: I guess you need to dodge names (spanNameFromResponse not spanName) because the request and response parser methods are in the same type and you can't overload?

jcchavezs: Exactly because of that.

src/Zipkin/Instrumentation/Http/Server/DefaultHttpServerParser.php
```diff
@@ -32,16 +32,41 @@ protected function spanName(Request $request): string
...
+        return $response->getRequest()->getMethod() . ' ' . $response->getRoute();
+    }
+
+
     /**
      * {@inhertidoc}
      */
     public function response(Response $response, TraceContext $context, SpanCustomizer $span): void
     {
+        $spanName = $this->spanNameFromResponse($response);
```
jcchavezs, line 65: I think this is the most important piece of logic @adriancole .

adriancole: k so I guess you can choose to do things like this, or switch to separate parsers. when brave switched it untangled things like this a fair amount, especially when response always has a ref to the request

jcchavezs: Yes I saw it (you pointed out this in the past) but I think I will keep this cohesive logic for now to keep it simple. Also these methods aren't exposed (protected) so the impact of this is controllable.

PR #185, opened 2020-10-21
tests: adds support for the rout e in the HTTP response.

jcchavezs:
Currently the base test was not comparing the response route.


tests/Unit/Instrumentation/Http/Server/BaseResponseTest.php
```diff
@@ -25,6 +26,15 @@ abstract class BaseResponseTest extends TestCase
         string $route = null
     ): array;
 
+    /**
+     * supportsRoute tells if the request implementation supports storing the
+     * route or not.
+     */
+    protected static function supportsRoute(): bool
```
jcchavezs, line 33: the reason why this is not abstract is to not to be a breaking change.

PR #191, opened 2020-11-21
GitHub actions

peter279k:
For PR #190 fix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,4 @@
-# GitHub Action for Symfony
-name: Testing Symfony
+name: CI
 on: [push, pull_request]
 jobs:
   test:
@@ -33,9 +32,22 @@ jobs:
       - name: Install Composer dependencies
         run: |
           composer install --no-progress --prefer-dist --optimize-autoloader
+      - name: Download coveralls.phar
+        if: matrix.operating-system != 'windows-latest'
+        run: |
+          wget -c -nc --retry-connrefused --tries=0 https://github.com/satooshi/php-coveralls/releases/download/v1.0.1/coveralls.phar
+          chmod +x coveralls.phar
+          php coveralls.phar --version
+          mkdir -p build/logs
       - name: Run Lint
+        if: matrix.operating-system != 'windows-latest'
         run: composer lint
       - name: Run Static check
         run: composer static-check
       - name: Run Tests
-        run: composer test
+        run: composer test -- --coverage-clover=build/logs/clover.xml
+      - name: Upload coverage report to codecov service
+        if: matrix.operating-system != 'windows-latest'
+        run: |
+          php coveralls.phar -v
+          bash <(curl -s https://codecov.io/bash)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,41 @@
+# GitHub Action for Symfony
+name: Testing Symfony
+on: [push, pull_request]
+jobs:
+  test:
+    name: Zipkin PHP (PHP version ${{ matrix.php-versions }} on ${{ matrix.operating-system }})
+    runs-on: ${{ matrix.operating-system }}
+    strategy:
+      fail-fast: false
+      matrix:
+        operating-system: [ubuntu-latest, windows-latest, macos-latest]
+        php-versions: ['7.1', '7.2', '7.3', '7.4']
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup PHP, with composer and extensions
+        uses: shivammathur/setup-php@v2 #https://github.com/shivammathur/setup-php
+        with:
+          php-version: ${{ matrix.php-versions }}
+          extensions: mbstring, xml, ctype, iconv, intl, pdo_sqlite
+          coverage: xdebug #optional
+      - name: Get composer cache directory
+        id: composer-cache
+        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+      - name: Cache composer dependencies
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,42 @@
...
+          path: ${{ steps.composer-cache.outputs.dir }}
+          # Use composer.json for key, if composer.lock is not committed.
+          # key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
+          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+          restore-keys: ${{ runner.os }}-composer-
+      - name: Install Composer dependencies
+        run: |
+          composer install --no-progress --prefer-dist --optimize-autoloader
+      - name: Run Lint
+        if: matrix.operating-system != 'windows-latest'
```
peter279k, line 37: Ignoring the `composer lint` on Windows operating system.

tests/Unit/Reporters/JsonV2SerializerTest.php
```diff
@@ -87,6 +87,13 @@ public function testJSONTagsAreSerializedCorrectly()
             . '"duration":1000,"localEndpoint":{"serviceName":"service1"},'
             . '"tags":{"test_key_1":"{\"name\":\"Kurt\"}","test_key_2":"foo\nbar"}'
             . '}]';
+        if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') {
```
peter279k, line 90: The expected should use `\r\n` for new line on Windows.

tests/Integration/Reporters/Http/CurlFactoryTest.php
```diff
@@ -14,6 +14,13 @@
  */
 final class CurlFactoryTest extends TestCase
 {
+    protected function setUp(): void
+    {
+        if (extension_loaded('pcntl') === false) {
```
peter279k, line 19: Checking the `pcntl` extension is loaded because the `pcntl` extension is only available on Unix-based operating system.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,42 @@
+# GitHub Action for Symfony
```
jcchavezs, line 1:
```suggestion
```

tests/Unit/Reporters/JsonV2SerializerTest.php
jcchavezs: Maybe instead of duplicating this code we can add a constant for new line? Actually that would be much better.

jcchavezs:
Thanks a lot for this. If have time, do you mind also migrating the code coverage part as in https://github.com/openzipkin/zipkin-php/blob/master/.travis.yml

peter279k:
> Thanks a lot for this. If have time, do you mind also migrating the code coverage part as in https://github.com/openzipkin/zipkin-php/blob/master/.travis.yml

I will  add that.

tests/Unit/Reporters/JsonV2SerializerTest.php
peter279k: Using the `str_replace` function to avoid the duplicated codes.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,54 @@
...
+          restore-keys: ${{ runner.os }}-composer-
+      - name: Install Composer dependencies
+        run: |
+          composer install --no-progress --prefer-dist --optimize-autoloader
+      - name: Download coveralls.phar
+        if: matrix.operating-system != 'windows-latest'
+        run: |
+          wget -c -nc --retry-connrefused --tries=0 https://github.com/satooshi/php-coveralls/releases/download/v1.0.1/coveralls.phar
+          chmod +x coveralls.phar
+          php coveralls.phar --version
```
peter279k, line 41: The coverage will be uploaded on `macos-latest` and `ubuntu-latest` operating systems now.

.github/workflows/ci.yml
peter279k: Removed.

peter279k:
@jcchavezs, it seems that the `coveralls` need to have additional setting on GitHub Action.

And the `codecov` service has been uploaded to the remote server.

tests/Unit/Reporters/JsonV2SerializerTest.php
```diff
@@ -87,6 +87,9 @@ public function testJSONTagsAreSerializedCorrectly()
...
+            $expectedSerialization = str_replace('\\n', '\\r\\n', $expectedSerialization);
```
jcchavezs, line 91: I am curious about this @adriancole: when a tag contains linebreaks and when we escape them and turn linebreaks into `\n` and for Linux or MacOS, what do we do for windows, do we turn into `\n` too or `\r\n`?

peter279k: When I look at the PHPUnit testing result, it seems that this assertion will be failed on Windows due to the line-breaks issue.

jcchavezs:
Thanks a lot for this!

peter279k:
> Thanks a lot for this!

Thanks for your time to concern this PR!

tests/Unit/Reporters/JsonV2SerializerTest.php
adriancole: most unit tests don't include line breaks, as tags are sent opaquely and there are no standard tags that have linebreaks in them anyway.. avoid the problem instead?

peter279k: It can use the `PHP_EOL` constraint value defined on internal PHP and more details are available on PR #193.

PR #200, opened 2021-01-26
Upgraded middlewares/fast-route and middlewares/request-handler to their 2.0.0 versions in order to support PHP 8.0.

FutureMatt:
So it looks like support for PHP 7.1 would need to be dropped with this as the `middleware` packages do not support 7.1. But given that it has been EOL for the last year this may be a good thing. We'd need to bump `"php": "^7.1||^8.0"` to `"php": "^7.2||^8.0"` in `composer.json`.


jcchavezs:
@FutureMatt how about accepting both versions first? Wouldn't that still working?

andrei-dascalu:
Just an observation: it would be great to have this stepping stone before v3 but:

1. not sure it's ideal to drop a PHP version without a major version bump
2. In my previous MR for PHP8 support ended up dropping both 7.1 and 7.2. I started with this update (and dropping 7.1), which in turn led to having to replace guzzle psr7 by nyholm psr7 (and dropping 7.2). It's a bit of a snowball.

jcchavezs:
At the beginning we thought about supporting 8.0 without dropping 7.1 but it seems not feasible. I think it is safe to close this PR and keep working towards 3.0 to support 8.0. I am even thinking the same way we only supported 5.6 along with 7.x in 1.0 we probably only want to support 7.4 and all the 8.x.

jcchavezs:
Superseeded by https://github.com/openzipkin/zipkin-php/pull/203

Issue #202, opened 2021-05-26
Support for PHP 8.0


andrei-dascalu:
Hi,

I've noticed that there are 2 PR's already on this topic (#200 and #198 ), so I thought it might be ok to also have an issue about the challenges around this. Also being selfish and all I was hit by the incompatibility with PHP 8 as we're going through a migration process with a couple of projects that also use zipkin.

Given that both PHP 7.1 & PHP 7.2 are past their EOL dates, is there a support target/plan with respect to backwards compatibility of various PHP versions?

Thanks

jcchavezs:
Hi @andrei-dascalu thanks for creating this.

This is something desirable and I hope we can get something functional in the next couple of weeks. Current challenges are mainly on making tests pass. Main blocker is this error https://github.com/openzipkin/zipkin-php/pull/198#issuecomment-813097562

Regarding dropping support for 7.1 and 7.2 I think we can happily cut *3.0* including features only compatible with 7.3+. I could even think about 7.4+ and 8.0 only.

andrei-dascalu:
@jcchavezs I have a proposition about that, if it's ok. I'm going to make a merge request (looks ok in my local but there are some other changes there). What I found is that Guzzlehttp/psr7 doesn't provide PSR-17 factories, only psr7. There's some WIP there but so far in similar situations I found nyholm to be more usable in that it provides both.

andrei-dascalu:
Party on! Looking forward to v3.

jcchavezs:
Do you have in mind any features we can benefit from PHP 7.3+? I think
JSON_THROW_ON_ERROR and maybe properties type. Besides that I think a great
set of features could be added when 8.0 is landed (e.g. usage of
annotations for autotracing).

Do we need to support 7.3 or just go with 7.4 for v3.0?

Any thoughts @limingxinleo @FutureMatt?

limingxinleo:
I don't think php8 is popular yet, so it's necessary to support php7.3

limingxinleo:
Hi, @jcchavezs 

How long can you deal with this issue?

It's my pleasure to get your reply.

Thanks.

jcchavezs:
I hope we can close a release by end of this month.

nocive:
bump

jcchavezs:
[...]

PR #203, opened 2021-05-26
PHP 8: upgrade middlewares, phpunit and replace guzzle psr7 with nyholm

andrei-dascalu:
Proposal for issue #202 

- there was a bit of a snowball there with respect to aligning dependencies to work with minor PHP versions (I did not try too hard to find a way to keep 7.2 as per issue comments).
-  See https://github.com/php-http/discovery/issues/164 - guzzlehttp doesn't provide expected PSR-17 factories (there's code available ready for their v2, but that would mean dropping some stability constraints) => switched to nyholm (it claims some performance boost as well, though probably not too relevant since we're talking tests)
- some tweaks around the tests: eg `dechex()` needs an `int`, but `hexdec()` returns `int|float`, running afoul of strong type check.
- updated phpunit also due to version constraints (phpunit 8 was still complaining about running on php 8) which led to lots of warnings about prophecy having to use a trait for integration with phpunit (could ignore those I guess to aim for 7.2 compatibility), which requires `phpspec/prophecy-phpunit` which in turn requires min. 7.3
- rename class that did not conform to PSR-4

### DIFF
--- a/tests/Unit/Instrumentation/Http/Server/Psr15/MiddlewareTest.php
+++ b/tests/Unit/Instrumentation/Http/Server/Psr15/MiddlewareTest.php
@@ -19,9 +19,9 @@ use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\RequestInterface;
 use PHPUnit\Framework\TestCase;
-use GuzzleHttp\Psr7\ServerRequest;
+use Nyholm\Psr7\ServerRequest;
 
-final class ServerTest extends TestCase
+final class MiddlewareTest extends TestCase
 {
     private static function createTracing(callable $requestSampler = null): array
     {
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         operating-system: [ubuntu-latest, windows-latest, macos-latest]
-        php-versions: ["7.2", "7.3", "7.4", "8.0"]
+        php-versions: ["7.3", "7.4", "8.0"]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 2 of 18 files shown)


tests/Unit/Instrumentation/Http/Server/Psr15/MiddlewareTest.php
```diff
@@ -19,9 +19,9 @@
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\RequestInterface;
 use PHPUnit\Framework\TestCase;
-use GuzzleHttp\Psr7\ServerRequest;
+use Nyholm\Psr7\ServerRequest;
 
-final class ServerTest extends TestCase
+final class MiddlewareTest extends TestCase
```
jcchavezs, line 24: Nice catch.

jcchavezs:
Current failure is about

image

I added this to keep retrocompatibility with Zipkin PHP 1.x. I don't think we need to carry that into 3.x. Could you comment out that error in phpstan.neon and then I will drop that magic in a next PR?

andrei-dascalu:
@jcchavezs done!
I've also commented out the line about "# if openssl_random_pseudo_bytes we want to fail" as it seems that when the issue is not hit, it counts as an error for the static check. I wasn't able to get the error running with php8.

However, it does appear in PHP7.4, so pre-8 these two should be the other way around (`# if openssl_random_pseudo_bytes we want to fail` -> ignored and the added one `# This avoids false positive for parameter type mismatch` -> not ignore).

Is there a way to have neon files per php version ?

andrei-dascalu:
nevermind, added `reportUnmatchedIgnoredErrors: false`

andrei-dascalu:
I've made another correction to the neon file but at this time I'm shooting in the dark. Locally I tried under docker with php7.3 and 7.4 and it seems ok.

jcchavezs:
I think I will first remove the backwards support so it does not cause you issues in this PR.

jcchavezs:
I am OK with this, coveralls failed on this. I am merging as it is. Thank you!

PR #204, opened 2021-05-27
Proposed fix for coveralls

andrei-dascalu:
Seems that PHPUnit 9 doesn't accept the clover path via parameter, resulting in no file generated so I've added it via PHPUnit config. The git action spec already uses xdebug 3 for coverage so this should work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
       - name: Run static check
         run: composer static-check
       - name: Run tests
-        run: composer test -- --coverage-clover=build/logs/clover.xml
+        run: composer test -- 
       - name: Upload coverage report to codecov service
         if: matrix.operating-system == 'ubuntu-latest'
         run: |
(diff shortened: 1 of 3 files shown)


jcchavezs:
It is failling in ubuntu, is it possible this is because we run this as a fork?

andrei-dascalu:
@jcchavezs i guess it is possible. Looking back I see the same error that no repository was found. I am not sure how it's supposed to work. Have you seen this before? Is it possible to try to execute the CI manually?

jcchavezs:
The merge to master did work

image


So I guess this is failing because of this being a fork. I wonder if we still need this PR given that the merge did work.

andrei-dascalu:
I think it should work without, st least judging by the errors in the previous merge request. I will close this and keep the code on my side should it be needed later.

jcchavezs:
Do you have in mind any features we can benefit from PHP 7.3+? I think
JSON_THROW_ON_ERROR and maybe properties type. Besides that I think a great
set of features could be added when 8.0 is landed (e.g. usage of
annotations for autotracing).

José Carlos Chávez


tor. 27. mai 2021 kl. 20:59 skrev Andrei Dascalu ***@***.***>:

PR #206, opened 2021-06-09
Add Memcached Reporter

### LINKED ISSUE #22: Explore possibilities for bulk reporting
At the moment, every trace (a set of spans) from a request will end up being one http request to the reporter. This is not ideal but limitation in the language as PHP is, in many real-world contexts, single-threaded and does not have an event loop like Node.js so we must flush when the request finishes. This is something convenient at small scale but not much at a big scale. As the outcome of this issue we should come up with a solution to avoid the overhead on the immediate reporter vs. bulk reporter.

Ping @adriancole @basvanbeek @beberlei @felixfbecker

Clivern:
This PR adds memcached reporter. So basically it is similar to `InMemory` but it stores the value in a memcached server. It also should work as expected in case of concurrent reads and writes. it uses this feature from memcached https://www.php.net/manual/en/memcached.cas.php. This requires `php-memcached` extension, so it will be a breaking change. 

### Usage

In application, we use `Memcached` reporter not `HTTP` reporter.

```php
<?php
// app.php
include_once dirname(__FILE__) . "/vendor/autoload.php";

use Zipkin\Annotation;
use Zipkin\Endpoint;
use Zipkin\Samplers\BinarySampler;
use Zipkin\TracingBuilder;
use Zipkin\Reporters\Aggregation\MemcachedClient;
use Zipkin\Reporters\Memcached;

$endpoint = Endpoint::create('clivern');

$reporter = new Memcached([], new MemcachedClient('127.0.0.1'));

$sampler = BinarySampler::createAsAlwaysSample();
$tracing = TracingBuilder::create()
    ->havingLocalEndpoint($endpoint)
    ->havingSampler($sampler)
    ->havingReporter($reporter)
    ->build();

$tracer = $tracing->getTracer();

for ($i=1; $i <= 10000; $i++) {
  $span = $tracer->newTrace();
  $span->setName(sprintf("span_%d", $i));
  $span->start();

  try {
    // something quite fast
    var_dump("test");
  } finally {
    $span->finish();
  }

  $tracer->flush();
}
```

Then we create a cron job or a command to read from `Memcached` reporter and send to `Zipkin` using `HTTP` reporter. We may even add this as application middleware with a time interval.

```php
<?php
// daemon.php
include_once dirname(__FILE__) . "/vendor/autoload.php";

use Zipkin\Reporters\Aggregation\MemcachedClient;
use Zipkin\Reporters\Memcached;
use Zipkin\Reporters\Http;

$reporter = new Memcached([], new MemcachedClient('127.0.0.1'));
$httpReporter = new Http(['endpoint_url' => 'http://zipkin.local:9411/api/v2/spans']);

while (true) {
    $spans = $reporter->flush();

    if (!empty($spans)){
        $httpReporter->report($spans);
    }

    sleep(5);
}
```

Flushing can be based on a time interval or the stored value size (Memcached support value up to 1MB by default that's ~1000000 chars). i didn't add any implementation since it is totally up to the developer and the scale zipkin is used. reporting 2 spans/second will have a different implementation than 1000 spans/second and I think this shouldn't be part of the package. in case you will hit the 1MB size limit in a second, you can either increase memcached maximum object size or report with different memcached keys, so basically you split the collected spans across different memcached keys to get more space before flushing.

Please let me know if something need to be adjusted!

Fix #22

### Update:

To run as single process

```php
<?php

include_once dirname(__FILE__) . "/vendor/autoload.php";

use Zipkin\Annotation;
use Zipkin\Endpoint;
use Zipkin\Samplers\BinarySampler;
use Zipkin\TracingBuilder;
use Zipkin\Reporters\Aggregation\MemcachedClient;
use Zipkin\Reporters\Memcached;
use Zipkin\Reporters\Http;


$endpoint = Endpoint::create('orders_service');

$httpReporter = new Http([
  'endpoint_url' => 'http://zipkin.local:9411/api/v2/spans'
]);

$reporter = new Memcached(
    [],
    $httpReporter,
    new MemcachedClient('127.0.0.1')
);

$sampler = BinarySampler::createAsAlwaysSample();
$tracing = TracingBuilder::create()
    ->havingLocalEndpoint($endpoint)
    ->havingSampler($sampler)
    ->havingReporter($reporter)
    ->build();

$tracer = $tracing->getTracer();

for ($i=1; $i <= 400; $i++) {
  $span = $tracer->newTrace();
  $span->setName(sprintf("span_%d", $i));
  $span->start();

  try {
    var_dump("report");
  } finally {
    $span->finish();
  }

  $tracer->flush();
}
```


src/Zipkin/Reporters/Http.php
```diff
@@ -82,7 +82,8 @@ public function report(array $spans): void
             return;
         }
 
-        $payload = $this->serializer->serialize($spans);
+        $payload = (is_object($spans[0])) ? $this->serializer->serialize($spans) : json_encode($spans);
```
jcchavezs, line 85: I am not clear about this change, what is the reason we need this?

composer.json
```diff
@@ -22,6 +22,7 @@
     "require": {
         "php": "^7.3 || ^8.0",
         "ext-curl": "*",
+        "ext-memcached": "*",
```
jcchavezs, line 25: I think we can leave this optional because you only need this if you are actually using the client. I think we can check if extension is loaded when constructing the client.

src/Zipkin/Reporters/Aggregation/MemcachedClient.php
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * @param string       $server
+     * @param int          $port
+     * @param bool         $enableCompression
+     */
+    public function __construct(
+        string $server = '127.0.0.1',
+        int $port = 11211,
+        bool $enableCompression = true
```
jcchavezs, line 37: I think this options can go in an array and be passed directly to ->setOption iteratively. Does that make sense?

src/Zipkin/Reporters/Memcached.php
```diff
@@ -0,0 +1,176 @@
...
+    /**
+     * @param  array  $spans
+     */
+    public function report(array $spans): void
+    {
+        try {
+            $this->memcachedClient->ping();
+
+            // Fetch stored spans
+            $result = $this->memcachedClient->get(
```
jcchavezs, line 68: I would move this right before `empty($result)` for the sake of consistency.

src/Zipkin/Reporters/Memcached.php
```diff
@@ -0,0 +1,176 @@
...
+                $this->memcachedClient->quit();
+                return;
+            }
+
+            $status = false;
+
+            // Merge the new spans with the stored spans only if
+            // the item not updated by a different concurrent proceess
+            while (!$status) {
+                $result['value'] = array_merge(
```
jcchavezs, line 94: I think decoding the payloads to merge is an overhead, we already know they are both `serialized JSON array of objects` so we can just do the operations manually. It is possible to create another interface that would play along with `JsonV2Serializer` to merge two payloads as I described.

src/Zipkin/Reporters/Memcached.php
```diff
@@ -0,0 +1,176 @@
...
+                );
+
+                $status = $this->memcachedClient->cas(
+                    $result['cas'],
+                    $this->options['cache_key'],
+                    json_encode($result['value'])
+                );
+
+                if (!$status) {
+                    $result = $this->memcachedClient->get(
```
jcchavezs, line 106: I think we can move this to the first line after the while and get rid of the previous `get`.

jcchavezs:
@Clivern thanks for starting this

> Then we create a cron job or a command to read from Memcached reporter and send to Zipkin using HTTP reporter. We may even add this as application middleware with a time interval.

I have a question regarding this, do we truly need such a daemon? I can think of: on reporting if the time has passed or if the max size has been reached we run the reporter. Does that sound reasonable?

Also we should consider the max size for reporting, even with the daemon we should know how big is the reporting payload because server might have a limit (e.g. in zipkin-go it is https://github.com/openzipkin/zipkin-go/blob/master/reporter/http/http.go#L37)

Clivern:
Thanks @jcchavezs for the review, i will check them case by case.

Clivern:
> @Clivern thanks for starting this
> 
> > Then we create a cron job or a command to read from Memcached reporter and send to Zipkin using HTTP reporter. We may even add this as application middleware with a time interval.
> 
> I have a question regarding this, do we truly need such a daemon? I can think of: on reporting if the time has passed or if the max size has been reached we run the reporter. Does that sound reasonable?
> 
> Also we should consider the max size for reporting, even with the daemon we should know how big is the reporting payload because server might have a limit (e.g. in zipkin-go it is https://github.com/openzipkin/zipkin-go/blob/master/reporter/http/http.go#L37)

It make sense, if you want to collect and send from one process. I tried to be flexible and also if you want to remove the daemon, you can simply run the above on each incoming request to the application, like middleware but still it is a lot to do indeed. So i can add that piece of logic too so memcached reporter will do the aggregation and reporting to zipkin if configurable size or time reached. only issue i see, when reporting time reached, reporting to zipkin will slow down things a bit but should be fine.

```php
include_once dirname(__FILE__) . "/vendor/autoload.php";

use Zipkin\Reporters\Aggregation\MemcachedClient;
use Zipkin\Reporters\Memcached;
use Zipkin\Reporters\Http;

$reporter = new Memcached([], new MemcachedClient('127.0.0.1'));
$httpReporter = new Http(['endpoint_url' => 'http://zipkin.local:9411/api/v2/spans']);

$last_reporting_time = /*call memcached here*/

if(time() - $last_reporting_time > 86400) {
    $spans = $reporter->flush();

    if (!empty($spans)){
        $httpReporter->report($spans);
    }
}
```

src/Zipkin/Reporters/Http.php
jcchavezs: I see. Thanks for clarifying this. Let me explore some alternatives.

jcchavezs:
So I think: 
1. We can store all traces reported in memcached serialized using `serialize`, join them on every report and flush them into HTTP after x seconds (`batch_interval`) or max number of traces accumulated (`batch_size`, we need to have a counter for the number of spans, if we decide to join manually the serialized data, it includes already a counter on third position `a:3:{i:0;O:1:`). When it is time to report, we `unserialize` all the chunks, join them together all the traces and send them over HTTP. **The reason for using `serialize`/`deserialize` over json_encode/json_decode** is because we can't turn a json_decoded array into `ReadbackSpan` whereas we can do it with `serialize`/`unserialize` without overhead, this is if we store it in json we need to create a method for turning the json_decoded into `Readbackspan` and just for the sake of this exporter which sounds not enough reason for introducing a breaking change in the interface.
2. We could provide both daemon and the reporting in single process, we just need to make sure in docs that `batch_interval=-1` and `batch_size=-1` will skip all reporting cc @basvanbeek. The daemon would do just the reporting and won't accept this options because the execution of a cron can't be controled by our PHP code.

Does that make sense?

Clivern:
> So I think:
> 
> 1. We can store all traces reported in memcached serialized using `serialize`, join them on every report and flush them into HTTP after x seconds (`batch_interval`) or max number of traces accumulated (`batch_size`, we need to have a counter for the number of spans, if we decide to join manually the serialized data, it includes already a counter on third position `a:3:{i:0;O:1:`). When it is time to report, we `unserialize` all the chunks, join them together all the traces and send them over HTTP. **The reason for using `serialize`/`deserialize` over json_encode/json_decode** is because we can't turn a json_decoded array into `ReadbackSpan` whereas we can do it with `serialize`/`unserialize` without overhead, this is if we store it in json we need to create a method for turning the json_decoded into `Readbackspan` and just for the sake of this exporter which sounds not enough reason for introducing a breaking change in the interface.
> 2. We could provide both daemon and the reporting in single process, we just need to make sure in docs that `batch_interval=-1` and `batch_size=-1` will skip all reporting cc @basvanbeek. The daemon would do just the reporting and won't accept this options because the execution of a cron can't be controled by our PHP code.
> 
> Does that make sense?

Yes, will apply the first point first (aggregate & report in a single process) & from there, we can see how it make it easy for everyone to opt-in and opt-out by changing `batch_interval` and `batch_size` values

src/Zipkin/Reporters/Aggregation/MemcachedClient.php
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * Compare and swap an item.
+     *
+     * @param float  $casToken
+     * @param string $key
+     * @param mixed  $value
+     * @param int    $expiration
+     */
+    public function cas($casToken, $key, $value, $expiration = 0): bool
```
jcchavezs, line 108: I know cas is the name in Memcached work but I think it would be really cool if we could name it `compareAndSwap`.

src/Zipkin/Reporters/Memcached.php
```diff
@@ -0,0 +1,245 @@
...
+                $status = $this->memcachedClient->cas(
+                    $result['cas'],
+                    sprintf("%s_spans", $this->options['cache_key_prefix']),
+                    serialize($result['value'])
+                );
+            }
+
+            $this->memcachedClient->quit();
+        } catch (Exception $e) {
+            $this->logger->error(
```
jcchavezs, line 124: I think if `(count($result['value']) >= $this->options['batch_size']))` and we fail we still need to empty batch import because imagine memcached is down, this is only growing and growing and retrying won't get any better.

jcchavezs:
This is great work @Clivern I just added a couple of comments then I am happy to merge it.

src/Zipkin/Reporters/Memcached.php
Clivern: give it a look now & tell me what do you think? so if memcached is down, i will send the current reported spans inside the catch call. aggregation will be ignored in that case since memcached is not functional. 

it shouldn't grow inside memcache once `batch_interval` or `batch_size` reached. it is always better to have both of them activated. i added values by default.

we update memcached at the last step and if it failed, we will report the current reported spans inside the catch call. and still the same amount of spans inside memcached. once memcached back online, still size will exceed and finally we can send whatever was there.

I will update the test cases, add timeout to memcached client and validate the behaviour for any flaws

Clivern:
> This is great work @Clivern I just added a couple of comments then I am happy to merge it.

@jcchavezs I added test cases and did a few changes & tested it manually. please review it again.

jcchavezs:
Great work @Clivern! I will merge it as soon as it gets green.

jcchavezs:
@Clivern could you run the linter please?

src/Zipkin/Reporters/Memcached.php
```diff
@@ -0,0 +1,252 @@
...
+    /**
+     * @param array           $options
+     * @param Reporter        $reporter
+     * @param MemcachedClient $memcachedClient
+     * @param LoggerInterface $logger
+     * @param SpanSerializer  $serializer
+     */
+    public function __construct(
+        array $options = [],
+        Reporter $reporter,
```
jcchavezs, line 57: I think reporter is mandatory so options can't have a default value.

src/Zipkin/Reporters/Aggregation/MemcachedClient.php
```diff
@@ -0,0 +1,131 @@
...
+
+        if (!class_exists('\Memcached')) {
+            throw new Exception("PHP ext-memcached is required");
+        }
+
+        $this->client = new \Memcached();
+        $this->client->addServer($this->server, $this->port);
+
+        $options = \array_merge(self::DEFAULT_OPTIONS, $options);
+        foreach($options as $key => $value) {
```
jcchavezs, line 61:
```suggestion
        foreach ($options as $key => $value) {
```

src/Zipkin/Reporters/Memcached.php
```diff
@@ -0,0 +1,252 @@
...
+
+    /**
+     * @param array           $options
+     * @param Reporter        $reporter
+     * @param MemcachedClient $memcachedClient
+     * @param LoggerInterface $logger
+     * @param SpanSerializer  $serializer
+     */
+    public function __construct(
+        array $options = [],
```
jcchavezs, line 56:
```suggestion
        array $options,
```

Clivern:
> Great work @Clivern! I will merge it as soon as it gets green.

it passes now locally, just trigger it one more time.

```
$ zipkin-php git:(master) composer static-check
> phpstan analyse src --level 8
Note: Using configuration file /Users/Ahmed/Space/Personal/oss/zipkin-php/phpstan.neon.
 72/72 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%


 [OK] No errors
```

jcchavezs:
do you mind rebasing, I just added this https://github.com/openzipkin/zipkin-php/pull/207 which I think will fix the problem. If this does not fix it I will merge this PR as it is.

Clivern:
done

Clivern:
it seems the github action `shivammathur/setup-php` doesn't install memcached extension even it is there.

jcchavezs:
Maybe we try specifing memcached explicitly?
https://github.com/shivammathur/setup-php/issues/423#issuecomment-799882630
José Carlos Chávez


ons. 30. jun. 2021 kl. 12:17 skrev Ahmed ***@***.***>:

Clivern:
nope! it doesn't work, i tested it. they don't mention windows at all, all about ubuntu. not surprised! should be hard to get things right on windows

only workaround i can think of right now is to add the following to top of test cases in https://github.com/openzipkin/zipkin-php/blob/master/tests/Unit/Reporters/MemcachedTest.php. anyways sdk must work even if memcached not there and repository ubuntu workflow works fine with memcached extension.

```php
if (!class_exists('\Memcached')) {
        $this->markTestSkipped('Memcached extension is missing.');
        return;
}
```
```zsh
composer test
Do not run Composer as root/super user! See https://getcomposer.org/root for details
> phpunit tests
PHPUnit 9.5.6 by Sebastian Bergmann and contributors.

...............................................................  63 / 210 ( 30%)
............................................................... 126 / 210 ( 60%)
.....................SSSSSSSSSSSSSS............................ 189 / 210 ( 90%)
.....................                                           210 / 210 (100%)

Time: 00:00.172, Memory: 12.00 MB

OK, but incomplete, skipped, or risky tests!
Tests: 210, Assertions: 447, Skipped: 14.
```
in that case we will have to skip these two files from phpstan analysis by adding these to end of `phpstan.neon`. didn't find a better way so far to bypass "not found memcached class" error if class not exist.

```yaml
    excludePaths:
        analyse:
            - src/Zipkin/Reporters/Aggregation/MemcachedClient.php
            - src/Zipkin/Reporters/Memcached.php
```

jcchavezs:
Sounds like a good tradeoff for now. Do you ming opening a PR for that?

José Carlos Chávez


ons. 30. jun. 2021 kl. 15:18 skrev Ahmed ***@***.***>:

jcchavezs:
I moved this code temporary to a branch, I need to release 3.0 and I would like to do some tweaks to the implementation but I don't want to block 3.0 so this is most likely included in 3.1.

PR #223, opened 2022-01-03
Fixes coverage

jcchavezs:
This PR fixes the coverage value as currently it is misleading.

The reason for this coverage value is that we were running 
```
 "test": [
            "@test-unit",
            "@test-integration"
        ],
```

So first target `test-unit` generated a report and then `test-integration` was overriding it and hence giving a false value.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -66,10 +66,7 @@
     "scripts": {
         "fix-lint": "phpcbf --standard=ZEND --standard=PSR2 --ignore=*/vendor/* ./",
         "lint": "phpcs --standard=ZEND --standard=PSR2 --ignore=*/vendor/* ./",
-        "test": [
-            "@test-unit",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -66,10 +66,7 @@
     "scripts": {
         "fix-lint": "phpcbf --standard=ZEND --standard=PSR2 --ignore=*/vendor/* ./",
         "lint": "phpcs --standard=ZEND --standard=PSR2 --ignore=*/vendor/* ./",
-        "test": [
-            "@test-unit",
-            "@test-integration"
-        ],
+        "test": "phpunit tests/Unit",
```
jcchavezs, line 69:
```suggestion
        "test": "phpunit tests",
```

jcchavezs:
image

PR #225, opened 2022-05-13
 Zipkin\\RealSpan::tag() must be of the type string

Tinywan:
```
Argument 2 passed to Zipkin\\RealSpan::tag() must be of the type string, int given
```


jcchavezs:
Thank you!

